PR #48, opened 2021-10-08
Feature/always close body

kstiehl:
### Description

It is recommended to always close the response.Body since otherwise connections could be dangling. However the current code was not doing it properly when an 5xx or 6xx error was returned the functions exited early and never closed the body

### Release Note

```release-note
* general - fix a bug where connections could end up dangling around
```

### Community Note

### DIFF
--- a/pkg/ipam/address/address.go
+++ b/pkg/ipam/address/address.go
@@ -172,7 +172,7 @@ func (a api) Get(ctx context.Context, id string) (Address, error) {
 	return responsePayload, err
 }
 
-func (a api) Delete(ctx context.Context, id string) (err error) {
+func (a api) Delete(ctx context.Context, id string) error {
 	url := fmt.Sprintf(
 		"%s%s/%s",
 		a.client.BaseURL(),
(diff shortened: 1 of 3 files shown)


pkg/ipam/address/address.go
```diff
@@ -187,11 +189,15 @@ func (a api) Delete(ctx context.Context, id string) error {
...
+	defer func() {
+		err = httpResponse.Body.Close()
+	}()
+
 	if httpResponse.StatusCode >= 500 && httpResponse.StatusCode < 600 {
 		return fmt.Errorf("could not execute address delete request, got response %s", httpResponse.Status)
 	}
 
-	return httpResponse.Body.Close()
+	return nil
```
LittleFox94, line 200: `err` is initialized to `nil`, making the explicit `nil` redundant - but no idea if having it explicit is better or not.

LittleFox94:
Most of the time, the (maybe) returned error from `httpResponse.Body.Close()` is ignored, despite the functions having an error return - is this something we should change, to always return the error from Close() if it exists - at least when we don't have another error already?

pkg/ipam/address/address.go
LittleFox94: I think I'd prefer not having it explicit, in case someone changes the function in the future, adding a branch where `err` is set but not returning right away for some reason and the return at the end throws away the error because of this

PR #49, opened 2021-10-08
✨: Implement endpoints to get locations by identifier and code

toothstone:
### Description

This adds wrapper routines for the core/location/`<identifier>` and core/location/by-code/`<code>`

### Release Note

```release-note
ENHANCEMENTS
* core/location - Add getters for locations by identifier and code
```

### DIFF
--- a/pkg/core/location/api.go
+++ b/pkg/core/location/api.go
@@ -6,9 +6,11 @@ import (
 	"github.com/anexia-it/go-anxcloud/pkg/client"
 )
 
-// API contains methods for location listing.
+// API contains methods for retrieving and listing locations.
 type API interface {
 	List(ctx context.Context, page, limit int, search string) ([]Location, error)
+	Get(ctx context.Context, identifier string) (Location, error)
+	GetByCode(ctx context.Context, code string) (Location, error)
 }
 
 type api struct {
--- a/pkg/core/location/location.go
+++ b/pkg/core/location/location.go
@@ -10,7 +10,7 @@ import (
 )
 
 const (
-	pathPrefix = "/api/core/v1/location.json"
+	pathPrefix = "/api/core/v1/location"
 )
 
 // Location is the metadata of a single location.
@@ -33,7 +33,7 @@ type listResponse struct {
 
 func (a api) List(ctx context.Context, page, limit int, search string) ([]Location, error) {
 	url := fmt.Sprintf(
-		"%s%s?page=%d&limit=%d&search=%s",
+		"%s%s.json?page=%d&limit=%d&search=%s",
 		a.client.BaseURL(),
 		pathPrefix, page, limit, search,
 	)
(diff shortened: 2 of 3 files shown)


pkg/core/location/location.go
```diff
@@ -60,3 +60,65 @@ func (a api) List(ctx context.Context, page, limit int, search string) ([]Locati
...
+	if err != nil {
+		return Location{}, fmt.Errorf("could not execute location get-by-code request: %w", err)
+	}
+	if httpResponse.StatusCode >= 500 && httpResponse.StatusCode < 600 {
+		return Location{}, fmt.Errorf("could not execute location get-by-code request, got response %s", httpResponse.Status)
+	}
+
+	var location Location
+	err = json.NewDecoder(httpResponse.Body).Decode(&location)
+	_ = httpResponse.Body.Close()
```
LittleFox94, line 118: use a `defer httpResponse.Body.Close()` right after the error check for `a.client.Do()` - look at #48 for lots of examples

tests/core_test.go
```diff
@@ -26,11 +26,24 @@ var _ = Describe("Core API endpoint tests", func() {
 
 	Context("Location endpoint", func() {
 
-		It("Should list all available locations", func() {
+		It("Should list all available locations, and get the first entry by ID and code", func() {
```
LittleFox94, line 29: Better have this at three separate tests imho, `It("Should list all available locations", ...)` and `It("Should get the first entry by ID", ...)` and `It("Should get the first entry by code", ...)`

pkg/core/location/api.go
```diff
@@ -6,9 +6,11 @@ import (
 	"github.com/anexia-it/go-anxcloud/pkg/client"
 )
 
-// API contains methods for location listing.
+// API contains methods for getting and listing location.
```
LittleFox94, line 9: While `getting` clearly communicated the HTTP verb, I think `retrieving` sounds a bit better. Also, in this sentence,`locations` would be better than `location`.

pkg/core/location/api.go
```diff
@@ -6,9 +6,11 @@ import (
...
 type API interface {
 	List(ctx context.Context, page, limit int, search string) ([]Location, error)
+	Get(ctx context.Context, identifier string) (Location, error)
```
LittleFox94, line 12: I'd prefer `GetByID` instead

pkg/core/location/location.go
```diff
@@ -60,3 +60,65 @@ func (a api) List(ctx context.Context, page, limit int, search string) ([]Locati
...
+	if err != nil {
+		return Location{}, fmt.Errorf("could not execute location get request: %w", err)
+	}
+	if httpResponse.StatusCode >= 500 && httpResponse.StatusCode < 600 {
+		return Location{}, fmt.Errorf("could not execute location get request, got response %s", httpResponse.Status)
+	}
+
+	var location Location
+	err = json.NewDecoder(httpResponse.Body).Decode(&location)
+	_ = httpResponse.Body.Close()
```
LittleFox94, line 87: same as already commented below: use `defer`

pkg/core/location/location.go
toothstone: Done via 346a0c3

pkg/core/location/api.go
toothstone: Done via cc1f6a6

LittleFox94: Only the first sentence though ;)

pkg/core/location/api.go
LittleFox94: Had the wrong reference point then, as I looked at the LBaaS package for reference - would love a comment from @kstiehl for this, if he says it's fine as it is, this is ok-to-merge for me

tests/core_test.go
```diff
@@ -26,13 +26,30 @@ var _ = Describe("Core API endpoint tests", func() {
...
+		It("Should get the first location entry by ID", func() {
+			ctx, cancel := context.WithTimeout(context.Background(), 3*time.Minute)
+			defer cancel()
+			list, _ := location.NewAPI(cli).List(ctx, 1, 1000, "")
+			l, err := location.NewAPI(cli).Get(ctx, list[0].ID)
+			Expect(err).NotTo(HaveOccurred())
+			Expect(l).To(Equal(list[0]))
+		})
+
+		It("Should get the first location entry by ID", func() {
```
LittleFox94, line 45: Copy-Paste error - Code instead of ID

pkg/core/location/api.go
kstiehl: I think it's very hard to compare these two packages. The API design on Anexia side is very different. 

@toothstone is right that almost everywhere in go-anxcloud Get implies that an ID is used. However this is not the case for the lbaas package. Why is that so? Because the LBAAS API is very different to all the other APIs. I think the main difference is that executing a get without specifying an ID is that you will receive a list of `xxxInfo` Objects in difference to when you execute it with an ID you receive the actual `xxxx` Object (where `xxx` is either `Frontend`, `Backend`, `Server`) 

For most of the other APIs this is not the case. That's why we have `Get` and `GetByID` seperated in the lbaas package.

I think this PR is fine. If we really want to unify things then we should probably change it in the lbaas package. which comes with the cost of hiding the semantic difference between the `Get` and `GetByID` endpoint.  

I don't really have a strong opinion about that. I can see advantages on both sides

tests/core_test.go
```diff
@@ -26,13 +26,30 @@ var _ = Describe("Core API endpoint tests", func() {
...
 			ctx, cancel := context.WithTimeout(context.Background(), 3*time.Minute)
 			defer cancel()
 			_, err := location.NewAPI(cli).List(ctx, 1, 1000, "")
 			Expect(err).NotTo(HaveOccurred())
 		})
 
+		It("Should get the first location entry by ID", func() {
+			ctx, cancel := context.WithTimeout(context.Background(), 3*time.Minute)
+			defer cancel()
+			list, _ := location.NewAPI(cli).List(ctx, 1, 1000, "")
```
kstiehl, line 39:
```suggestion
			list, err := location.NewAPI(cli).List(ctx, 1, 1000, "")
			Expect(err).NotTo(HaveOccurred())
```

pkg/core/location/api.go
LittleFox94: Hm.. I'd think the LBaaS `Get()` is the same as `List()` in all other APIs, since that name is only important in this library. We should probably change that in LBaaS.

Nevertheless, this PR is fine then.

PR #64, opened 2021-10-29
🎨✅💚: refactor tests for CloudDNS/Zone

LittleFox94:
### Description

Tests are now moved into the package, can be run with mocks instead of the real Engine (which is the default behavior, pass the command line flag "-integration-tests" to the tests to run them against the live Engine API) and, most importantly, generate a random zone name to not conflict with other tests against the Engine in parallel.

Also done:
* `make func-test` now also generates the HTML coverage report as `make test` already does 
* extracted `pkg/utils/test.RandomHostname()` from `vsphere` tests to also use in `clouddns` tests

### Release Note

```release-note
NONE
```

### References

References all the failing integration tests.

### Community Note


toothstone:
Hi, I can't seem to get the CloudDNS tests to run in integration test mode, `make func-tests` actually has them passing even if there is no ANEXIA_TOKEN set. Am I holding it wrong? I think it would be helpful to put the exact commands into the README here, we can't expect people to read the Makefile to find out about some feature flag.

Also the phrasing of the PR and commit description is somewhat misleading, to me it reads as if the `--integration-tests` flag is activating the mocks.

Lastly, as a general note, there is quite a lot of stuff happening within this one PR, which makes it hard to review. Going forward it would be nice to have separate feature PRs (new mocks) and refactoring PRs (move to pkg, introduce randomness, generalize hostname generator) where possible, or at least some more dedicated commits.

LittleFox94:
> `make func-tests` actually has them passing even if there is no ANEXIA_TOKEN set

hm this is expected to fire against the real API, will look into it - thanks o.o

> Also the phrasing of the PR and commit description is somewhat misleading, to me it reads as if the `--integration-tests` flag is activating the mocks.

Will change

> Lastly, as a general note, there is quite a lot of stuff happening within this one PR, which makes it hard to review. Going forward it would be nice to have separate feature PRs (new mocks) and refactoring PRs (move to pkg, introduce randomness, generalize hostname generator) where possible, or at least some more dedicated commits.

Hm true .. sorry, will take more care in the future

LittleFox94:
There was indeed a bug in running the CloudDNS against the real Engine - this is fixed now and fails as intended without `ANEXIA_TOKEN` set in the environment.

LittleFox94:
The coverage decrease is due to updating a newly creating zone failing with a server error (ENGSUP-5233):
https://github.com/anexia-it/go-anxcloud/pull/64/files#diff-ee83c858f2de755dae95084e93f0416bd100c0180c8b1a85a92dfdf903150d7aR223

Makefile
```diff
@@ -33,7 +33,16 @@ test:
 
 .PHONY: func-test
 func-test:
-	CGO_ENABLED=1 go test -coverpkg ./pkg/... -coverprofile coverage.out -timeout 180m ./tests/...
+	# we run the tests in our tests directory and for every package listed additionally
+	# some of the newer tests (starting with the refactored CloudDNS tests) can run as
+	# either unit or integration tests, using mocks of the API for unit tests. They need
+	# the --integration-tests flag to run as integration tests.
```
toothstone, line 39: flag looks different below

toothstone (changes requested):

Makefile
LittleFox94: I wanted to point out that go does not differentiate between `--integration-tests` and `-integration-tests`, but the flag is not actually plural - good catch x)

toothstone (approved):
LGTM

We need a general overhaul of our testing instruction, I've created https://ats.anexia-it.com/browse/SYSENG-682 to track that. To keep things moving, I'm fine with merging this PR as it is now.

LittleFox94:
Thanks for writing that issue :)

Merging now

PR #71, opened 2021-11-05
✨: allow Object identifiers in embedded fields

LittleFox94:
### Description

This extends `getObjectIdentifier` to recursively search for identifier field in embedded/anonymous fields, too.

### Release Note

```release-note
NONE (addition to unreleased feature)
```

### References

Generic client introduced in #56

### Community Note

### DIFF
--- a/pkg/api/object.go
+++ b/pkg/api/object.go
@@ -22,8 +22,12 @@ func getObjectIdentifier(obj types.Object, singleObjectOperation bool) (string, 
 }
 
 func findIdentifierInStruct(t reflect.Type, v reflect.Value, singleObjectOp bool) (string, error) {
+	// we also use this to track if we found an identifier already
+	var returnIdentifier *string
+
 	numFields := t.NumField()
 
+fields:
 	for i := 0; i < numFields; i++ {
 		field := t.Field(i)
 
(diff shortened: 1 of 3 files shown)


pkg/api/object.go
```diff
@@ -18,19 +18,49 @@ func getObjectIdentifier(obj types.Object, singleObjectOperation bool) (string,
...
+			embeddedType := field.Type
+			embeddedValue := v.Field(i)
+
+			for embeddedType.Kind() == reflect.Ptr {
+				embeddedType = embeddedType.Elem()
+				embeddedValue = embeddedValue.Elem()
+			}
+
+			if embeddedType.Kind() == reflect.Struct {
+				if ret, err := findIdentifierInStruct(embeddedType, embeddedValue, singleObjectOp); err == nil {
```
toothstone, line 44: Just out of curiosity: Is this DoS-able by having a (malicious) self-referencing or circular struct definition?

LittleFox94: Plain structs embedded in itself does not compile: https://play.golang.org/p/AM5EbV_OG3m

Struct embedding a pointer to itself compiles and would indeed be an endless loop here, but this can only be done by authors or users of this package - not by data returned from the engine or anything else at runtime, so I think this is only a theoretical problem not worth the complexity of checking that.

toothstone (approved):

PR #72, opened 2021-11-05
SYSENG-686 ✨: Introduce ResponseDecodeHook for manipulating returned objects

toothstone:
### Description

This adds the ResponseDecodeHook as discussed in SYSENG-686.

Using this hook, an Object can have completely custom decoding, matching the already existing `RequestBodyHook`. The Object is required to decode the data in the given `io.Reader` into the receiver, storing the decoded data into itself.

To implement this with non-redundant code, the logic to decode an Object is exported into it's own function, used by all parts of the generic client (`api_implementation.go/doRequest()`, `api_implementation.go/List()/channel goroutine function` and `pagination.go/Next()`).

While implementing this new hook, we realized many arguments passed to the hook functions are actually kind of "the context". Because of this, we extended the context used by the generic client to include the `Operation`, `Options` and, after `EndpointURL` was called, the original URL. We plan to change the existing Hook functions to grab this information from the context instead of arguments.
Using a context allows us to add more to it, without having to change all the existing hooks.

This is a team effort of @toothstone and @LittleFox94 :muscle: 

### References

* #56 introduction of generic API client

### Release Note

```release-note
NONE (additions to unreleased code)
```

### DIFF
--- a/pkg/api/api_implementation.go
+++ b/pkg/api/api_implementation.go
@@ -6,6 +6,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
+	"io/ioutil"
 	"mime"
 	"net/http"
 	"net/url"
@@ -212,6 +213,10 @@ func (a defaultAPI) List(ctx context.Context, o types.FilterObject, opts ...type
 					// scoped variables makes the data for the closure perfectly identified.
 					closureData := o
 					c <- func(out types.Object) error {
+						if decodeResponse, ok := out.(types.ResponseDecodeHook); ok {
+							return decodeResponse.DecodeAPIResponse(&closureData, req.URL, types.OperationList, &options)
+						}
+
 						return json.Unmarshal(closureData, out)
 					}
 				}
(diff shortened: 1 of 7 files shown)


pkg/api/api_implementation.go
```diff
@@ -439,3 +440,23 @@ func addPaginationQueryParameters(req *http.Request, opts types.ListOptions) {
 
 	req.URL.RawQuery = query.Encode()
 }
+
+func decodeResponse(ctx context.Context, mediaType string, data io.Reader, res interface{}) error {
+	if mediaType == "application/json" {
+		if decodeResponse, ok := res.(types.ResponseDecodeHook); ok {
```
LittleFox94, line 446: one of the iterations included a check for `if not a List operation`. We don't actually need it, since List operations decode into a `[]json.RawMessage`, which never is a `types.ResponseDecodeHook`

toothstone: Good catch...
I'll just pretend I did this on purpose, since my life is of course always guided by PEP20.

On a more serious note, maybe some documentation how these routines interact with the different methods might be helpful for future development.

toothstone:
Ca. tomorrow I'll rebase/adapt my new CloudDNS pkg to the new interface, if that works we can move this forward.

pkg/api/api_implementation.go
LittleFox94: "Flat is better than nested."? :stuck_out_tongue: 

More docs would definitely be helpful. Do you have any specific suggestions what to document? Having a hard time finding _what_ to document .. writing the docs isn't that big of a problem.

toothstone:
> Ca. tomorrow I'll rebase/adapt my new CloudDNS pkg to the new interface, if that works we can move this forward.

I've rebased and implemented the feature this hook was originally meant for, which works fine. I'm happy with the interface as it is now.

PR #74, opened 2021-11-12
Generic client: handle empty response bodies

LittleFox94:
### Description

This PR allows the generic client to handle empty/missing response bodies.

### Release Note
```release-note
NONE (fix for unreleased generic client)
```

### Community Note

### DIFF
--- a/pkg/api/api_implementation.go
+++ b/pkg/api/api_implementation.go
@@ -362,11 +362,15 @@ func (a defaultAPI) doRequest(req *http.Request, obj types.Object, body interfac
 		return err
 	}
 
-	if mediaType, err := getResponseType(response); err == nil {
-		return decodeResponse(req.Context(), mediaType, response.Body, body)
-	} else {
-		return err
+	if response.StatusCode != http.StatusNoContent {
+		if mediaType, err := getResponseType(response); err == nil {
+			return decodeResponse(req.Context(), mediaType, response.Body, body)
+		} else {
+			return err
+		}
 	}
+
+	return nil
 }
 
 func (a defaultAPI) contextPrepare(ctx context.Context, o types.Object, op types.Operation, opts types.Options) (context.Context, error) {
--- a/pkg/api/api_test.go
+++ b/pkg/api/api_test.go
@@ -401,7 +401,7 @@ var _ = Describe("creating API with different options", func() {
 	})
 
 	It("handles the Engine returning a weird response content-type", func() {
-		server.AppendHandlers(ghttp.RespondWith(200, nil, http.Header{"Content-Type": []string{"application/octet-stream"}}))
+		server.AppendHandlers(ghttp.RespondWith(200, "randomgarbage", http.Header{"Content-Type": []string{"application/octet-stream"}}))
 
 		api, err := NewAPI(
 			WithClientOptions(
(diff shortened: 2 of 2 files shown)


toothstone:
I have some second thoughts on this whole topic...

- It might cause problems down the line if we choose to simply ignore empty (faulty!) application/json responses on the SDK level. If another Engine API breaks and starts to respond with such garbage, users (and we) should get alerted about it by having go-anxcloud err out
- The broken behavior this works around is (so far) confined to one Engine service, i.e. package. Shouldn't it be handled within that package, encapsulated from the rest of go-anxcloud?

I think we're between a rock and a hard place here, not sure what to do about it.

pkg/api/api_implementation.go
```diff
@@ -359,16 +359,25 @@ func (a defaultAPI) doRequest(req *http.Request, obj types.Object, body interfac
...
-
-		// unreachable, getResponseType() already checks for supported types
-		return ErrUnsupportedResponseFormat
-	} else {
+	emptyResponse, responseBodyReader, err := checkEmptyReader(response.Body)
+	if err != nil {
 		return err
 	}
+
+	if !emptyResponse {
```
toothstone, line 367: I think it is kind of the wrong way around if we first check the body, and afterwards the media type - shouldn't we first look at the media type, and then based on that decide what to do with the content?
There could be other mediaTypes where an empty body is completely unacceptable and we need to return an error.

LittleFox94:
> * It might cause problems down the line if we choose to simply ignore empty (faulty!) application/json responses on the SDK level. If another Engine API breaks and starts to respond with such garbage, users (and we) should get alerted about it by having go-anxcloud err out
> * The broken behavior this works around is (so far) confined to one Engine service, i.e. package. Shouldn't it be handled within that package, encapsulated from the rest of go-anxcloud?

I think at least `204 No Content` responses are perfectly valid and no garbage, especially as response to a `DELETE` request.

In a first iteration of this fix, I checked for `Content-Length == 0 || StatusCode == 204`, but that would leave cases broken where the golang http client does not know the Content-Length (chunked transfer for example) - I'm not sure how relevant that is for the Engine.

Accepting `emptyResponse` for every method might be bad, maybe we should at least error for `Get` and `List`.

pkg/api/api_implementation.go
LittleFox94: sadly the engine sends a content-type of application/json together with a `204 No Content` .. so that does not work

toothstone:
> I think at least `204 No Content` responses are perfectly valid and no garbage, especially as response to a `DELETE` request.

Could we just check if the response is a 204, and if so, stop decoding? In that case there is no need to check the body size or mediaType at all, afaik.

LittleFox94:
> Could we just check if the response is a 204, and if so, stop decoding? In that case there is no need to check the body size or mediaType at all, afaik.

Yes, but I'm pretty sure (not knowing, gut feeling) the Engine will send responses without body with `200 OK` ^^'

toothstone:
> > Could we just check if the response is a 204, and if so, stop decoding? In that case there is no need to check the body size or mediaType at all, afaik.
> 
> Yes, but I'm pretty sure (not knowing, gut feeling) the Engine will send responses without body with `200 OK` ^^'

I propose we take care of that once it actually comes up^^
For now not decoding 204 is sufficient and seems completely valid to me.

toothstone:
Please rebase, will then cherry-pick it for my CloudDNS proof of concept where it is needed.

toothstone (approved):
Looks great to me, I'm very happy we found a standard-conforming solution here.

PR #79, opened 2021-11-18
Add `FullObjects` option, retrieving full objects on listing

LittleFox94:
### Description

I think title says enough already ^^'

This also fixes some other small things, neatly in their own commits.

### Release Note
Release note for [CHANGELOG](https://github.com/anexia-it/go-anxcloud/blob/master/CHANGELOG.md):

```release-note
* NONE (updating unreleased code)
```

### References
* generic client introduced in #56
* SYSENG-684

### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

PR #81, opened 2021-11-22
✨💥: fundamentally change List via channel

LittleFox94:
### Description

Previous to this commit, we let users create a channel variable and pass
a pointer to this to the AsObjectChannel option, which initialized it
and made the API client spawn a goroutine to fill it.

We now let the user create a "ObjectChannelCloser" via NewObjectChannel
and pass that to the (renamed) ObjectChannel option. Via this new
ObjectChannelCloser, the user can retrieve the go channel to stream
objects as previous, but also cancel the listing - which was not
possible before and leaked goroutines whenever the channel was not read
until it finished.

### Release Note

```release-note
NONE (changing unreleased feature)
```

### References
* generic client introduced in #56
* found the missing feature while writing another test for #79 

### Community Note

### DIFF
--- a/pkg/api/object_channel.go
+++ b/pkg/api/object_channel.go
@@ -42,46 +42,7 @@ func (oc *objectChannel) Channel() (types.ObjectChannel, error) {
 	}
 
 	pageRetrieveWaiter := make(chan bool)
-
-	go func() {
-		var pageData []json.RawMessage
-
-	cancelChannel:
-		for oc.pageIterator.Next(&pageData) {
-			for _, o := range pageData {
-				select {
-				case <-oc.abort:
-					oc.log.Info("User requested aborting the channel")
-					break cancelChannel
-				case <-pageRetrieveWaiter:
-					oc.log.Info("Object retrieved by user")
-				}
-
-				oc.log.Info("Pushing retriever on channel")
-
-				// since we are in a goroutine, we might already be in the next iteration of this loop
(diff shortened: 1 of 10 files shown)


pkg/api/object_channel.go
```diff
@@ -0,0 +1,122 @@
...
+func NewObjectChannel() types.ObjectChannelCloser {
+	return &objectChannel{
+		objects: make(chan types.ObjectRetriever),
+		abort:   make(chan bool),
+	}
+}
+
+// Channel starts the object listing logic and returns the channel objects can be retrieved from.
+func (oc *objectChannel) Channel() (types.ObjectChannel, error) {
+	if oc.pageIterator == nil {
```
LittleFox94, line 40: I'm not entirely sure if we want to have such a check. It makes sense to block using a channel just created with NewObjectChannel(), but might make it unnecessarily hard to use it in goroutines (like one goroutine initializing and listing it, while another makes the `List()` call - because queuing API calls or whatever?). Not sure if this is a relevant use case.

kstiehl: I think this check is perfectly fine. When using this with goroutines you would probably just handover the `chan` and not the `objectChannel` object, since the Close function should be called from a coordinating higher goroutine.

kstiehl (approved):

LittleFox94:
Thanks for approval, will add some more tests since test coverage isn't that great yet (and found some bugs even)

PR #82, opened 2021-11-23
SYSENG-678 ✨: Implement clouddns/zone API with generic client

toothstone:
### Description

This adds a new package for parts of the Anexia CloudDNS zone management
API based on the generic client.

### Release Note

```release-note
ADDED
* New Anexia CloudDNS zone management API implementation based on the generic client

```


LittleFox94:
I'm waiting for #86 and will adapt this to that new location

pkg/apis/clouddns/v1/record_types.go
```diff
@@ -0,0 +1,14 @@
+package v1
+
+// anxcloud:object:hooks=ResponseDecodeHook,PaginationSupportHook
+
+type Record struct {
+	Identifier string `json:"identifier,omitempty" anxcloud:"identifier"`
+	ZoneName   string
```
LittleFox94, line 7: no json tag, do you mean not to encode this?
it needs a `json:"-"` tag then

pkg/apis/clouddns/v1/zone_genclient.go
```diff
@@ -0,0 +1,87 @@
...
+
+	return req, nil
+}
+
+func (z *Zone) FilterAPIRequestBody(ctx context.Context) (interface{}, error) {
+	op, err := types.OperationFromContext(ctx)
+	if err != nil {
+		return nil, err
+	}
+	// The Create and Update endpoints expect the Zone's name to be in the request body under the key "zoneName"
```
LittleFox94, line 46: the `name` attribute isn't expected by the API for Create and Update and we probably shouldn't send it, right?

pkg/apis/clouddns/v1/zone_types.go
```diff
@@ -0,0 +1,74 @@
...
+	Serial     int       `json:"serial"`
+	State      string    `json:"state"`
+}
+
+type DNSServer struct {
+	// Required - DNS Server name (FQDN).
+	Server string `json:"server"`
+
+	// DNS Server alias
+	Alias string
```
LittleFox94, line 19: is this supposed to be json-encoded as `Alias`?

LittleFox94 (changes requested):
Found some minor things I will happily fix once you confirm them :)

All in all it looks good

pkg/apis/clouddns/v1/record_types.go
toothstone: Indeed

pkg/apis/clouddns/v1/zone_genclient.go
toothstone: Yes, we should be strict in what we send

pkg/apis/clouddns/v1/zone_types.go
toothstone: Yes. Was actually already missed in the legacy client code...

toothstone:
I've addressed the issues you pointed out, thanks for that :)

Feel free to squash and merge as you see fit.

pkg/apis/clouddns/v1/zone_genclient.go
```diff
@@ -0,0 +1,91 @@
...
+	if op == types.OperationCreate || op == types.OperationUpdate {
+		zWithZoneName := struct {
+			Zone
+			ZoneName string `json:"zoneName"`
+		}{*z, z.Name}
+
+		// `name` does not exist as a field on the Engine API for these requests,
+		// so we strip it from the request body.
+		zWithZoneName.Name = ""
```
LittleFox94, line 55: `Name` is not tagged with `json:"omitempty"`, so we still send it - but empty, correct?

I'll try and fix if that is the case, intention is clear now :)

LittleFox94:
Want to check my two added commits @toothstone? Please merge when they lgty :)

LittleFox94 (approved):

pkg/apis/clouddns/v1/zone_genclient.go
toothstone: `Zone.Name` is tagged as `json:"name,omitempty" anxcloud:"identifier"`, and I had a look at the resulting JSON string, which did not contain the name field.

pkg/apis/clouddns/v1/zone_genclient.go
```diff
@@ -47,13 +47,12 @@ func (z *Zone) FilterAPIRequestBody(ctx context.Context) (interface{}, error) {
 	if op == types.OperationCreate || op == types.OperationUpdate {
 		zWithZoneName := struct {
 			Zone
-			ZoneName string `json:"zoneName"`
+			// for these operations the Engine expects the name to be in zoneName instead of name
+			// first we had the additional ZoneName attribute, but we can just rename the name attribute
+			// instead.
+			Name string `json:"zoneName"`
```
toothstone, line 53: I checked the resulting JSON with `json.NewEncoder(os.Stdout).Encode(zWithZoneName)` and it still contains a name field - so I think this re-annotating is not replacing the existing annotation, but extending it.

pkg/apis/clouddns/v1/zone_genclient.go
LittleFox94: alright, no idea what I looked at then - will roll my changes back and merge the rest :)

pkg/apis/clouddns/v1/zone_genclient.go
LittleFox94: yeah, you are right - this way it adds a field and not replaces it

PR #84, opened 2021-11-25
Code generator generating tests for Objects and optional hook interfaces

LittleFox94:
### Description

This adds a small code generator looking for magic comments in go sources. First version of this only knows some things about Objects (things the generic client can handle) and generates tests if they really implement the interfaces they intend to - especially the optional hook interfaces.

Code is expected to be generated by developers whenever they change something marked with a magic comment, a magic comment itself or the code generator. A CI job is in place to validate nothing is changed when generating code in CI - meaning everything is in git as expected.

### Release Note

```release-note
* NONE (dev change)
```

### References
* generic client introduced in #56 

### Community Note


kstiehl (approved):
From my perspective this looks good. 
One thing though. using markers in comments is a thing that is done by the go tooling itself and many frameworks. i.e `+build` comment. however the go tooling switched the naming convention to have a  namespace and a marker and discard the `+`.  so go `+build` became `go:build`. Currently many tools start to migrate to be compliant to this, in fact my IDE corrects some old markers to use the new ones.  Does it make sense to think of a similar marker scheme? `anxcloud:traits`

LittleFox94:
> From my perspective this looks good. One thing though. using markers in comments is a thing that is done by the go tooling itself and many frameworks. i.e `+build` comment. however the go tooling switched the naming convention to have a namespace and a marker and discard the `+`. so go `+build` became `go:build`. Currently many tools start to migrate to be compliant to this, in fact my IDE corrects some old markers to use the new ones. Does it make sense to think of a similar marker scheme? `anxcloud:traits`

I actually didn't notice they are also dropping the `+` in the new format but `anxcloud;` is meant as our namespacing prefix. At first I had `//+anxcloud:` defined as prefix, but since `specs` are separated by `;`, I changed to that as it looks more consistent.

Right now would be the best moment to change this - so, do we want to?

kstiehl:
> Another possible format would be separating specs by : and having key=value pairs as specs, instead of key:value:
// anxcloud:object:hooks=ResponseDecodeHook,RequestBodyHook

I like this one very much

kstiehl:
sorry one thing that just came to my mind. The advantage of using a pattern like `anxcloud:xxxx` would be that in the future we would be able to add more markers for other stuff (not sure if this will ever happen)

LittleFox94:
> sorry one thing that just came to my mind. The advantage of using a pattern like `anxcloud:xxxx` would be that in the future we would be able to add more markers for other stuff (not sure if this will ever happen)

with the proposed format we already can add more stuff - maybe we'd have to only use the `hooks` spec as it is defined now when the `object` spec is set, too, but other than that the format would allow that

kstiehl:
ACK.

kstiehl (approved):

PR #88, opened 2021-11-29
Fix template build sorting in tests

kstiehl:
### Description

For proper version sorting we need to extract the number from the `build` property of a template. The `build` Property is currently not enforcing any format however there is at least some kind of convention that I tried to put in a regex.

### Release Note

```release-note
None 
```

### Community Note

### DIFF
--- a/tests/vsphere_test.go
+++ b/tests/vsphere_test.go
@@ -45,17 +45,16 @@ const (
 var templateID string
 
 // This versioning scheme that currently seems to be in place for template build numbers.
-var buildNumberRegex = regexp.MustCompile(`[bB]?(?P<version>\d+)`)
+var buildNumberRegex = regexp.MustCompile(`[bB]?(\d+)`)
 
-func getBuildNumberFromString(build string) int {
-	index := buildNumberRegex.SubexpIndex("build")
+func extractBuildNumber(build string) int {
 	matches := buildNumberRegex.FindStringSubmatch(build)
 	if len(matches) == 0 {
 		// panic here since someone needs to check on the regex
 		panic("build does not match the buildNumberRegex")
 	}
 
-	number, err := strconv.ParseInt(matches[index], 10, 0)
+	number, err := strconv.ParseInt(matches[0], 10, 0)
 	if err != nil {
 		panic(fmt.Sprintf("could not extract build for %s: %s", build, err.Error()))
 	}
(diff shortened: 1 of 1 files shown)


tests/vsphere_test.go
```diff
@@ -43,6 +44,24 @@ const (
 
 var templateID string
 
+// This versioning scheme that currently seems to be in place for template build numbers.
+var buildNumberRegex = regexp.MustCompile(`[bB]?(?P<version>\d+)`)
```
LittleFox94, line 48: You declare a subexpression named `version` and some lines below get the index for `build`.

I'd just use this regex: `[bB]?(\d+)` and use `matches[0]` below.

Good job making me learn something about regex (named subexpression) though, that didn't happen in a while :D

tests/vsphere_test.go
```diff
@@ -43,6 +44,24 @@ const (
...
+
+func getBuildNumberFromString(build string) int {
```
LittleFox94, line 50: I'd call it `extractBuildNumber`, which is a bit shorter - but that's a big taste-only thing, am fine with this name too :)

LittleFox94 (changes requested):
I think I found a bug, please look at it again

tests/vsphere_test.go
kstiehl: oh. Forgot that during my last minute refactorings

tests/vsphere_test.go
kstiehl: No I was very unhappy with this name anyways, just didn't come up with a better one so far. So I will change this

tests/vsphere_test.go
```diff
@@ -45,17 +45,16 @@ const (
 var templateID string
 
 // This versioning scheme that currently seems to be in place for template build numbers.
-var buildNumberRegex = regexp.MustCompile(`[bB]?(?P<version>\d+)`)
+var buildNumberRegex = regexp.MustCompile(`[bB]?(d+)`)
```
LittleFox94, line 48: `\d`, not `d`

kstiehl: whoops

tests/vsphere_test.go
```diff
@@ -65,7 +83,7 @@ func vsphereTestInit() {
 	}
 
 	sort.Slice(selected, func(i, j int) bool {
-		return strings.Compare(selected[i].Build, selected[j].Build) > 0
+		return getBuildNumberFromString(selected[i].Build) > getBuildNumberFromString(selected[j].Build)
```
LittleFox94, line 86: you have to use the new name :)

LittleFox94 (changes requested):

tests/vsphere_test.go
kstiehl: i should really not do to things at the same time, it starts to get embarrassing :laughing:

PR #94, opened 2021-12-09
Test cleanup

LittleFox94:
### Description

This PR does a lot of things for test cleanup:
* moves the integration tests into their respective packages, getting rid of the `tests` directory in repository root.
  * tearing them apart in quite some cases
* cleaning up the integration test suites so they have a more useful structure of `Describe`, `Context` and `It`
* upgrades to ginkgo v2, so we can use `Ordered` containers (useful for e.g. `vsphere` tests)
* changes `make test` and `make func-test` to use `ginkgo` instead of `go test`

### Release Note

```release-note
NONE
```

### References
* [SYSENG-689](https://ats.anexia-it.com/browse/SYSENG-689)

### Community Note


LittleFox94:
I strongly suggest reviewing the individual commits, I took care to have meaningful ones. Reviewing the PR as whole is not going to be fun.

I also think "tests OK and coverage not dropped down" is a good review for most of the changes in this PR, but your view might be different :)

pkg/apis/clouddns/v1/integration_test.go
```diff
@@ -0,0 +1,6 @@
+//go:build integration
+// +build integration
+
+package v1
+
+const useMock = false
```
kstiehl, line 6: This is set to true when we are running integration tests. I think that we are using mocks here is just a consequence. So i would probably rename it to `isIntegrationTest` or something similar, since some tests wont even execute when this variable is set

kstiehl (approved):

pkg/apis/clouddns/v1/integration_test.go
LittleFox94: Good point, changed :)

PR #101, opened 2022-01-17
✅🩹: allow unknown fields in most responses

LittleFox94:
### Description

We configured the JSON decoder in use by the generic client to
`DisallowUnknownFields` before, because encoding/json does not know
about required fields, resulting in every json object being decodable
into every Go struct, but we want to make sure the response is actually
the type we expect.

Turns out this is only really needed for checking which style of
pagination response we got, not for objects and we already have
different code paths for decoding pagination responses and objects. This
commit allows unknown fields for objects.

We previously had a test case for a valid pagination response with
invalid objects in it. We did not see that in real life (as far as we
remember), so we removed that test case - hoping the Engine does not
throw such responses at us.

### Checklist

### References

* [SYSENG-753](https://ats.anexia-it.com/browse/SYSENG-753), not fully resolving the issue but doing the more urgent part of it

### Community Note

### DIFF
--- a/pkg/api/api_implementation.go
+++ b/pkg/api/api_implementation.go
@@ -483,11 +483,7 @@ func decodeResponse(ctx context.Context, mediaType string, data io.Reader, res i
 			return nil
 		}
 
-		d := json.NewDecoder(data)
-		// we have to disallow unknown fields since golangs json decoder does not have a way to specify required
-		// fields, meaning any json object can be decoded into any go struct.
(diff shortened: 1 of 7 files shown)


toothstone (approved):
lgtm, good job catching the customization in the CloudDNS client^^

LittleFox94:
> good job catching the customization in the CloudDNS client^^

another upside of code reviews: knowing which parts landed in `main` as workarounds :D

PR #107, opened 2022-02-03
Refactor and enhance generated Object tests

LittleFox94:
### Description

This PR refactors the generated Object tests to have most logic in a normal package which is imported by the generated code, rather than having the logic in the template string - making the generated tests much better to maintain.

Also adds two additional checks:
* test if the implemented hooks work fine with incomplete tests (needed for 100% test coverage, easy to test correctly with generated tests)
* test if the Object has an `identifier` field
  + we previously only had runtime errors for this

### Checklist

### References

* #84 

### Community Note


LittleFox94:
whoops, right, didn't implement the checks for all hooks, yet x)

pkg/utils/test/object.go
```diff
@@ -0,0 +1,119 @@
...
+			"ResponseFilterHook": func(ctx context.Context) error {
+				rec := httptest.NewRecorder()
+				rec.WriteHeader(200)
+				_, _ = rec.WriteString(`{}`)
+				_, err := o.(types.ResponseFilterHook).FilterAPIResponse(ctx, rec.Result())
+				return err
+			},
+		}
+
+		if errorCheck, ok := supportedHooks[name]; ok {
```
kstiehl, line 64: what if `hook` is not in the list of supportedHooks. Do we just not test it then? Or should we at least test whether the hook is actually implemented.

pkg/utils/test/object.go
```diff
@@ -0,0 +1,119 @@
...
+// * calling the hook function with incomplete contexts gives none or the correct error (meaning the error is handled correctly)
+//
+func ObjectTests(o types.Object, hooks ...interface{}) {
+	ginkgo.It("has an identifier", func() {
+		_, err := api.GetObjectIdentifier(o, false)
+		gomega.Expect(err).NotTo(gomega.HaveOccurred())
+	})
+
+	for _, hook := range hooks {
+		name := reflect.TypeOf(hook).Elem().Name()
```
kstiehl, line 32: I would move this line a few lines down. To initialize the variable shortly before it is used. makes it a little bit more readable

LittleFox94: ack

pkg/utils/test/object.go
LittleFox94: good catch, had code organized differently before where this made sense - will change

kstiehl (approved):
LGTM

PR #110, opened 2022-02-25
⬆️🔥: upgrade gh/satori/go.uuid and remove usages of it where possible

LittleFox94:
### Description

That library seems dead and we have to replace it, which would be a breaking change as it's part of the public interface of the old CloudDNS client.

For now, upgrading to latest master version fixes a security issue.

This PR also removes all usages not part of a public interface, so removing the whole dependency will be possible when the (already deprecated) old CloudDNS client is gone.

### Checklist

### References

* with SYSENG-1023 I planned to replace the UUID package, but after this PR merged and the already deprecated old clients removed, we can remove the dependency instead of replacing it

### Community Note

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,14 @@ e.g.
   - core/location
   - vlan
 
+### Changed
+* github.com/satori/go.uuid updated to latest master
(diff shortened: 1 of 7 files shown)


LittleFox94:
CI fails due to regression in CloudDNS API (ENGSUP-5697), but I think we can merge this anyway

PR #135, opened 2022-04-28
SYSENG-1230: fix missing LBaaS Backend attributes for Create operations

LittleFox94:
### Description

When creating a Backend, only a certain subset of attributes was sent to the Engine.

Also improves the request body handling for all the LBaaS objects, unifying Create and Update and omitting the State attribute instead of choosing the correct one to send and adds some `omitempty` tags to Object fields.

### Checklist

* [x] added release notes to `Unreleased` section in [CHANGELOG.md](CHANGELOG.md), if user facing change

### References
* [SYSENG-1230](https://ats.anexia-it.com/browse/SYSENG-1230)

### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

PR #136, opened 2022-05-02
SYSENG-1216: Refactor LBaaS E2E tests

LittleFox94:
### Description

This PR completely rewrites the LBaaS E2E tests to create less resources and actually test end-to-end (even with check if valid config is reachable on the test LBaaS server and fails with invalid config).

The old-style LBaaS client E2E tests are now labeled with `"old client", "slow"` and the Makefile was adapted to filter tests labeled like that, this way we can still run them if needed but skip by default.

### Checklist

### References
* builds on top of #135, only merge after that is merged

### Community Note


toothstone:
I'm maybe a little to the party, but is it really sustainable to do full e2e tests in the scope of an API client? I'm afraid this will set a dangerous precedent/expectation level and cost us a lot of resources to maintain.

LittleFox94:
> I'm maybe a little to the party, but is it really sustainable to do full e2e tests in the scope of an API client? I'm afraid this will set a dangerous precedent/expectation level and cost us a lot of resources to maintain.

If you are asking because of the "check if HA-Proxy is configured correctly"-tests: those are very little amount of code and I think a useful check, given the known problems with LBaaS.

As for having complete E2E tests in the API client at all: same for others, we also create VMs and stuff and given the known API problems, this again is probably a good idea. It shouldn't be the job of go-anxcloud, agreed, but better here than nowhere at all :/

toothstone:
> > I'm maybe a little to the party, but is it really sustainable to do full e2e tests in the scope of an API client? I'm afraid this will set a dangerous precedent/expectation level and cost us a lot of resources to maintain.
> 
> If you are asking because of the "check if HA-Proxy is configured correctly"-tests: those are very little amount of code and I think a useful check, given the known problems with LBaaS.

We'll have to run and maintain an LBaaS instance, for starters. So it is not just a little bit of code^^' That's a lot of idling resources and overhead to verify a handful of HTTP requests per day, if that.

> As for having complete E2E tests in the API client at all: same for others, we also create VMs and stuff and given the known API problems, this again is probably a good idea. It shouldn't be the job of go-anxcloud, agreed, but better here than nowhere at all :/

We create VMs, yes - but we don't set up SSH to log into them and check whether their disks have been sized as requested, which would be the equivalent here. IMHO we should limit the scope here so that tests only interact with the API we're wrapping.
We can't do QA for every dependency we have, so again this is a precedent I would rather not set.

Maybe we can move this discussion to tomorrow's Open Space? IMHO it's a team decision how much assurance we need vs what are we willing to spend on it.

LittleFox94:
yeah let's discuss this tomorrow, VM maintenance is a good point

pkg/apis/lbaas/v1/e2e_helpers_test.go
```diff
@@ -0,0 +1,161 @@
...
+}
+
+func waitObjectReady(ctx *context.Context, o *types.Object) {
+	waitObject(ctx, "eventually is ready", o, func(g Gomega, err error) {
+		Expect(err).NotTo(HaveOccurred())
+
+		hasState, ok := (*o).(StateRetriever)
+		Expect(ok).To(BeTrue())
+
+		Expect(hasState.StateFailure()).To(BeFalse())
```
marioreggiori, line 43: `g` namespaced `Expect` isn't used consistently. not sure if intentional

toothstone (approved):
We'll keep the e2e tests due to the recurring problems with LBaaS, but this is a clear exception and in no way expected or desirable for other APIs we wrap.
There will be further work to improve our test scheduling and to lower friction in the dev process.

pkg/apis/lbaas/v1/e2e_helpers_test.go
marioreggiori: besides that: lgtm

LittleFox94: it's intentional, but lacks comments about it - will add some

LittleFox94: adding the comments actually uncovered a wrong usage :upside_down_face: 

Every review is a good review!

marioreggiori (approved):

PR #122, opened 2022-04-12
Added helper functions for tagging and untagging of resources

marioreggiori:
### Checklist

* [x] added release notes to `Unreleased` section in [CHANGELOG.md](CHANGELOG.md), if user facing change

### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

PR #139, opened 2022-05-11
Implement generic mock client

marioreggiori:
### Checklist

### Community Note

### DIFF
--- a/pkg/api/mock/internal/core_types.go
+++ b/pkg/api/mock/internal/core_types.go
@@ -1,0 +1,30 @@
+package internal
+
+import (
+	"go.anx.io/go-anxcloud/pkg/api/types"
+	clouddnsv1 "go.anx.io/go-anxcloud/pkg/apis/clouddns/v1"
+	corev1 "go.anx.io/go-anxcloud/pkg/apis/core/v1"
+	lbaasv1 "go.anx.io/go-anxcloud/pkg/apis/lbaas/v1"
+	vlanv1 "go.anx.io/go-anxcloud/pkg/apis/vlan/v1"
+)
+
+// ObjectCoreType maps types.Object to *corev1.Type
+// returns nil if no mapping was found
+// TODO: (SYSENG-1333) have generic client bindings register themselves with their Type Identifier so we can use that for other purposes, too
+func ObjectCoreType(o types.Object) *corev1.Type {
+	switch o.(type) {
+	case *lbaasv1.Backend:
+		return &corev1.Type{Identifier: "33164a3066a04a52be43c607f0c5dd8c"}
+	case *lbaasv1.Bind:
+		return &corev1.Type{Identifier: "bd24def982aa478fb3352cb5f49aab47"}
+	case *lbaasv1.Frontend:
+		return &corev1.Type{Identifier: "da9d14b9d95840c08213de67f9cee6e2"}
+	case *lbaasv1.Server:
+		return &corev1.Type{Identifier: "01f321a4875446409d7d8469503a905f"}
+	case *vlanv1.VLAN:
+		return &corev1.Type{Identifier: "cf8e4dac56894afaa3244f6911bb62be"}
+	case *clouddnsv1.Zone:
+		return &corev1.Type{Identifier: "9190d73d8f4f42b5ad29e1a057f184fc"}
+	}
+	return nil
+}
--- a/pkg/api/mock/matcher/example_test.go
+++ b/pkg/api/mock/matcher/example_test.go
@@ -1,0 +1,102 @@
+package matcher
+
+import (
+	"context"
+	"fmt"
+	"log"
+
+	clouddnsv1 "go.anx.io/go-anxcloud/pkg/apis/clouddns/v1"
+	corev1 "go.anx.io/go-anxcloud/pkg/apis/core/v1"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	"go.anx.io/go-anxcloud/pkg/api/mock"
+)
+
+// Object is a custom gomega matcher which is used to test for matching objects
+// on the MockAPI instance by comparing the type and provided field names
+func ExampleObject() {
+	// defer GinkgoRecover() is not needed when assertion is done within Ginkgo environment
+	defer GinkgoRecover()
+
+	a := mock.NewMockAPI()
+
+	if err := a.Create(context.TODO(), &clouddnsv1.Record{ZoneName: "example.com", Name: "webmail"}); err != nil {
+		log.Fatalf("failed to create DNS record: %s", err)
+	}
+
+	// a.Existing() retrieves all objects which aren't currently flagged as destroyed
+	Expect(a.Existing()).To(
+		ContainElement(
+			Object(&clouddnsv1.Record{ZoneName: "example.com"}, "ZoneName"),
+		),
+	)
+
+	fmt.Println("Success")
+
+	// Output: Success
+}
+
+// Destroyed is a custom gomega matcher which is used to test
+// if an object was destroyed
+func ExampleDestroyed() {
+	// defer GinkgoRecover() is not needed when assertion is done within Ginkgo environment
+	defer GinkgoRecover()
+
+	a := mock.NewMockAPI()
+
+	record := clouddnsv1.Record{ZoneName: "example.com", Name: "webmail"}
+
+	if err := a.Create(context.TODO(), &record); err != nil {
+		log.Fatalf("failed to create record: %s", err)
+	}
+
+	if err := a.Destroy(context.TODO(), &clouddnsv1.Record{Identifier: record.Identifier}); err != nil {
+		log.Fatalf("failed destroy record: %s", err)
+	}
+
+	Expect(a.All()).To(
+		ContainElement(
+			SatisfyAll(
+				Destroyed(1),
+				Object(&clouddnsv1.Record{Identifier: record.Identifier}, "Identifier"),
+			),
+		),
+	)
+
+	fmt.Println("Success")
+
+	// Output: Success
+}
+
+// TaggedWith is a custom gomega matcher which is used to test
+// if an object was tagged with the provided tags
+func ExampleTaggedWith() {
+	// defer GinkgoRecover() is not needed when assertion is done within Ginkgo environment
+	defer GinkgoRecover()
+
+	a := mock.NewMockAPI()
+
+	zone := clouddnsv1.Zone{Name: "example.com"}
+
+	if err := a.Create(context.TODO(), &zone); err != nil {
+		log.Fatalf("failed to create DNS zone: %s", err)
+	}
+
+	if err := corev1.Tag(context.Background(), a, &zone, "Tag-0", "Tag-1", "Tag-2"); err != nil {
+		log.Fatalf("failed to tag zone: %s", err)
+	}
+
+	Expect(a.Existing()).To(
+		ContainElement(
+			SatisfyAll(
+				Object(&zone, "Name"),
+				TaggedWith("Tag-0", "Tag-1"),
+			),
+		),
+	)
+
+	fmt.Println("Success")
+
+	// Output: Success
+}
--- a/pkg/api/mock/matcher/matcher_object.go
+++ b/pkg/api/mock/matcher/matcher_object.go
@@ -1,0 +1,84 @@
+package matcher
+
+import (
+	"errors"
+	"fmt"
+	"reflect"
+	"strings"
+
+	"github.com/onsi/gomega/format"
+	gomegaTypes "github.com/onsi/gomega/types"
+	"go.anx.io/go-anxcloud/pkg/api/types"
+	"go.anx.io/go-anxcloud/pkg/utils/object/compare"
+)
+
+// Object matches APIObject with provided Object
+// It compares Objects by matchFields
+func Object(obj types.Object, matchFields ...string) gomegaTypes.GomegaMatcher {
+	return &objectMatcher{
+		obj:         obj,
+		matchFields: matchFields,
+	}
+}
+
+type objectMatcher struct {
+	obj         types.Object
+	matchFields []string
+}
+
+var (
+	// ErrObjectMatcherTypeMismatch is returned when ObjectMatcher is called with different Object types
+	ErrObjectMatcherTypeMismatch = errors.New("Provided Objects have different types")
+)
+
+func (m *objectMatcher) Match(actual interface{}) (bool, error) {
+	diff, err := m.compare(actual)
+	if err == ErrObjectMatcherTypeMismatch {
+		return false, nil
+	}
+	if err != nil {
+		return false, err
+	}
+
+	return len(diff) == 0, nil
+}
+
+func (m *objectMatcher) FailureMessage(actual interface{}) string {
+	diff, err := m.compare(actual)
+	if err != nil {
+		return err.Error()
+	}
+
+	formattedDiff := make([]string, 0, len(diff))
+	for _, d := range diff {
+		formattedDiff = append(formattedDiff, fmt.Sprintf(
+			"Key %q\n%s\n",
+			d.Key,
+			format.Message(d.A, "to equal", d.B),
+		))
+	}
+
+	return fmt.Sprintf("ACTUAL Object does not match EXPECTED Object:\n%s", strings.Join(formattedDiff, "\n"))
+}
+
+func (m *objectMatcher) NegatedFailureMessage(actual interface{}) string {
+	return "ACTUAL Object does match EXPECTED Object"
+}
+
+func (m *objectMatcher) compare(actual interface{}) ([]compare.Difference, error) {
+	obj, err := castToAPIObjectPointer(actual)
+	if err != nil {
+		return nil, err
+	}
+
+	if reflect.TypeOf(m.obj) != reflect.TypeOf(obj.Unwrap()) {
+		return nil, ErrObjectMatcherTypeMismatch
+	}
+
+	diff, err := compare.Compare(obj.Unwrap(), m.obj, m.matchFields...)
+	if err != nil {
+		return nil, err
+	}
+
+	return diff, nil
+}
--- a/pkg/api/mock/matcher/matcher_object_tags.go
+++ b/pkg/api/mock/matcher/matcher_object_tags.go
@@ -1,0 +1,63 @@
+package matcher
+
+import (
+	"fmt"
+
+	"github.com/onsi/gomega"
+	gomegaTypes "github.com/onsi/gomega/types"
+)
+
+// TaggedWith checks if APIObject is tagged with tags
+func TaggedWith(tags ...string) gomegaTypes.GomegaMatcher {
+	return &objectTaggedWith{
+		tags: tags,
+	}
+}
+
+type objectTaggedWith struct {
+	tags []string
+}
+
+func (m *objectTaggedWith) Match(actual interface{}) (bool, error) {
+	tags, err := extractObjectTags(actual)
+	if err != nil {
+		return false, err
+	}
+
+	return gomega.ContainElements(m.tags).Match(tags)
+}
+
+func (m *objectTaggedWith) FailureMessage(actual interface{}) string {
+	return m.commonFailureMessage(actual, false, "Object not tagged with all required tags")
+}
+
+func (m *objectTaggedWith) NegatedFailureMessage(actual interface{}) string {
+	return m.commonFailureMessage(actual, true, "Object tagged with unexpected tags")
+}
+
+func (m *objectTaggedWith) commonFailureMessage(actual interface{}, negated bool, failureMessage string) string {
+	tags, err := extractObjectTags(actual)
+	if err != nil {
+		return fmt.Sprintf("Failed to extract object tags: %s", err)
+	}
+
+	subMatch := gomega.ContainElements(m.tags)
+
+	var subFailureMessage string
+	if !negated {
+		subFailureMessage = subMatch.FailureMessage(tags)
+	} else {
+		subFailureMessage = subMatch.NegatedFailureMessage(tags)
+	}
+
+	return fmt.Sprintf("%s: %s", failureMessage, subFailureMessage)
+}
+
+func extractObjectTags(actual interface{}) ([]string, error) {
+	obj, err := castToAPIObjectPointer(actual)
+	if err != nil {
+		return nil, err
+	}
+
+	return obj.Tags(), nil
+}
--- a/pkg/api/mock/matcher/matcher_test.go
+++ b/pkg/api/mock/matcher/matcher_test.go
@@ -1,0 +1,205 @@
+package matcher
+
+import (
+	"context"
+	"fmt"
+	"testing"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	"github.com/onsi/gomega/types"
+	"go.anx.io/go-anxcloud/pkg/api/mock"
+	"go.anx.io/go-anxcloud/pkg/utils/object/compare"
+)
+
+func TestMockAPIMatcher(t *testing.T) {
+	RegisterFailHandler(Fail)
+	RunSpecs(t, "Mock API Matcher")
+}
+
+var _ = Describe("custom gomega matcher", func() {
+	var (
+		a       mock.API
+		objects []*testObject
+	)
+
+	BeforeEach(func() {
+		a = mock.NewMockAPI()
+
+		objects = make([]*testObject, 0, 8)
+
+		for i := 0; i < 4; i++ {
+			obj0 := testObject{TestFieldA: "text 0"}
+			obj1 := testObject{TestFieldA: "text 1"}
+			a.FakeExisting(&obj0)
+			a.FakeExisting(&obj1)
+			objects = append(objects, &obj0, &obj1)
+		}
+	})
+
+	Context("ContainElementNTimes", func() {
+		It("can test if the API list contains elements n times", func() {
+			Expect(a.Existing()).To(ContainElementNTimes(Object(&testObject{}), 8))
+			Expect(a.Existing()).To(ContainElementNTimes(Object(&testObject2{}), 0))
+		})
+
+		It("returns error when ACTUAL not of type array/slice/map", func() {
+			matcher := ContainElementNTimes(nil, 5)
+			_, err := matcher.Match(true)
+			Expect(err).To(HaveOccurred())
+			Expect(err.Error()).To(HavePrefix("ContainElementNTimes matcher expects an array/slice/map."))
+		})
+
+		It("fallbacks to equal matcher when element is no types.GomegaMatcher", func() {
+			matcher := ContainElementNTimes("test", 2)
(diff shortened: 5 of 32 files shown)


pkg/api/mock/mock_api_implementation_test.go
```diff
@@ -0,0 +1,208 @@
...
+						ObjectMatcher(&clouddnsv1.Record{ZoneName: "mock-2.go-anxcloud.test"}, "ZoneName"),
+						ObjectDeleted(),
+					),
+					2,
+				),
+			)
+		})
+	})
+
+	//Context("Object tagging", Ordered, func() {
```
marioreggiori, line 177: Tests depend on #122

pkg/api/mock/mock_api_implementation.go
```diff
@@ -0,0 +1,215 @@
...
+		val.Object = cloned.(types.Object)
+	} else {
+		return api.ErrNotFound
+	}
+
+	return nil
+}
+
+// Removes a types.Object from MockAPIs local storage
+func (a mockAPI) Destroy(ctx context.Context, o types.IdentifiedObject, opts ...types.DestroyOption) error {
```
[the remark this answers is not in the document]
marioreggiori: @LittleFox94 do you have any advice on how to solve this?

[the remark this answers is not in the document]
LittleFox94: having a `(a mockAPI) untag(o types.IdentifiedObject) (isTag bool, err error)` and call it like this would change two returns to only one:
```go
if isTag, err := a.untag(o); isTag {
  return err
}
```

something similar for `a.tag()` would help you for `Create`, too.

pkg/api/mock/mock_api_implementation.go
```diff
@@ -0,0 +1,215 @@
...
+			close(c)
+		}(channelPageIterator)
+	}
+
+	return nil
+}
+
+// Saves an types.Object to the MockAPIs local storage and sets a RandomIdentifier
+// if not already present. No validation is performed!
+func (a mockAPI) Create(ctx context.Context, o types.Object, opts ...types.CreateOption) error {
```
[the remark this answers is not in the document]
LittleFox94: See `Destroy` for an idea how to fix this

pkg/api/mock/mock_api_implementation_test.go
```diff
@@ -0,0 +1,226 @@
...
+
+		It("can Get resources", func() {
+			vlan := &vlanv1.VLAN{Identifier: testVLAN.Identifier}
+			err := a.Get(context.TODO(), vlan)
+			Expect(err).ToNot(HaveOccurred())
+			Expect(testVLAN).To(Equal(vlan))
+		})
+
+		It("can Update resources", func() {
+			copy, err := copystructure.Copy(testVLAN)
```
LittleFox94, line 64: I fear this is not a good test for this, this alternate one gives different results with real API vs mock API client:

```go
vlanUpdate := vlanv1.VLAN{
    Identifier: testVLAN.Identifier,
    Name:       "def",
}

err := a.Update(context.TODO(), &vlanUpdate)
Expect(err).ToNot(HaveOccurred())
Expect(vlanUpdate.Locations).To(Equal(testVLAN.Locations))
```

We mark a lot of fields with `omitempty`, resulting in them not being sent to the Engine, in turn resulting in those fields not being changed. This usage is an important one and mock client should behave as the real Engine API here

LittleFox94: Another detail in that example: Update requests most of the time get the full object in the response body, so `Locations` should be filled into `vlanUpdate` (as all the other attributes).

pkg/api/mock/mock_api_implementation_test.go
LittleFox94: Oh I see, we have a circular PR dependency

pkg/api/mock/mock_api_iterator.go
```diff
@@ -0,0 +1,74 @@
...
+
+import (
+	"errors"
+
+	"go.anx.io/go-anxcloud/pkg/api/types"
+)
+
+var (
+	// ErrPageSizeCannotBeZero is returned when newMockPageIter is getting called with itemsPerPage = 0 to prevent devision by zero panic
+	ErrPageSizeCannotBeZero = errors.New("mockPageIter itemsPerPage cannot be zero")
```
LittleFox94, line 11: hmm... what does the real Engine API do in such a case? :D

pkg/api/mock/mock_api_implementation_test.go
```diff
@@ -0,0 +1,226 @@
...
+)
+
+func TestMockAPI(t *testing.T) {
+	RegisterFailHandler(Fail)
+	RunSpecs(t, "Mock API tests")
+}
+
+var _ = Describe("Mock API implementation", func() {
+
+	Context("CRUD Operations", Ordered, func() {
```
LittleFox94, line 27: do we need them `Ordered`?

pkg/api/mock/matcher_object_tags.go
```diff
@@ -0,0 +1,40 @@
...
+	tags []string
+}
+
+func (m *objectTaggedWith) Match(actual interface{}) (bool, error) {
+	obj, ok := actual.(*mockAPIObject)
+	if !ok {
+		return false, fmt.Errorf("ObjectTaggedWith matcher expects an *mockAPIObject")
+	}
+
+	for _, tag := range m.tags {
```
LittleFox94, line 26: can you implement this with `ContainElements()`, using the `(Negated)FailureMessage` from that? Would give us a list of which tags are missing or wrongly set.

pkg/api/mock/matcher_object_deleted.go
```diff
@@ -0,0 +1,30 @@
+package mock
+
+import (
+	"fmt"
+
+	gomegaTypes "github.com/onsi/gomega/types"
+)
+
+// ObjectDeleted checks if mockAPIObject was deleted
+func ObjectDeleted(tags ...string) gomegaTypes.GomegaMatcher {
```
LittleFox94, line 10: ... what are the `tags` for?

pkg/api/mock/matcher_object_deleted.go
```diff
@@ -0,0 +1,30 @@
...
+	return &objectDeleted{}
+}
+
+type objectDeleted struct{}
+
+func (m *objectDeleted) Match(actual interface{}) (bool, error) {
+	obj, ok := actual.(*mockAPIObject)
+	if !ok {
+		return false, fmt.Errorf("ObjectDeleted matcher expects an *mockAPIObject")
```
LittleFox94, line 19: first: why not `errors.New()` instead of `fmt.Errorf()`

second: please don't ever return errors created on the fly ^^'

pkg/api/mock/matcher_object.go
```diff
@@ -0,0 +1,50 @@
...
+	diff, err := compare.Compare(obj.Object, m.obj, m.matchFields...)
+	if err != nil {
+		return false, err
+	}
+
+	return len(diff) == 0, nil
+}
+
+func (m *objectMatcher) FailureMessage(actual interface{}) string {
+	return "Matched Object is not present"
```
LittleFox94, line 45: first: this is a matcher for comparing a single ACTUAL Object with a single EXPECTED Object, so the message is more like "ACTUAL Object does not match EXPECTED Object"

second: having the difference printed would be very helpful, implementing this with an existing primitive matcher could help again

LittleFox94: hm re $second: using an existing primitive matcher is probably hard, but you can extract that info from `diff` returned by `compare.Compare`

that might mean executing `compare.Compare()` multiple times - for checking and for message, maybe you find a clever way of preventing that though

pkg/api/mock/matcher_object.go
```diff
@@ -0,0 +1,50 @@
...
+
+type objectMatcher struct {
+	obj         types.Object
+	matchFields []string
+}
+
+func (m *objectMatcher) Match(actual interface{}) (bool, error) {
+	obj, ok := actual.(*mockAPIObject)
+	if !ok {
+		return false, fmt.Errorf("ObjectMatcher matcher expects an *mockAPIObject")
```
LittleFox94, line 29: on-the-fly error again. and again `fmt.Errorf` without any placeholders in your format string ^^'

pkg/api/mock/mock_api_implementation.go
```diff
@@ -0,0 +1,241 @@
...
+			t.Field(i).Type.Kind() == reflect.String &&
+			v.Field(i).String() == "" {
+			v.Field(i).SetString(identifier)
+		}
+	}
+
+	identifier, err := api.GetObjectIdentifier(o, true)
+	if err != nil {
+		return err
+	}
```
LittleFox94, line 162: the block above generates and assigns an identifier and here you check if it has an identifier. This is probably a good test, but nothing to stay here in this runtime place

pkg/api/mock/mock_api_implementation.go
```diff
@@ -0,0 +1,241 @@
...
+}
+
+// Update overwrites a types.Object in MockAPIs local storage
+func (a mockAPI) Update(ctx context.Context, o types.IdentifiedObject, opts ...types.UpdateOption) error {
+	identifier, err := api.GetObjectIdentifier(o, true)
+	if err != nil {
+		return err
+	}
+
+	if val, ok := a[identifier]; ok && !val.Deleted {
```
LittleFox94, line 181: code style: please change to `!ok || val.Deleted` to return the error and have this, hopefully, default case outside the if

pkg/api/mock/mock_api_implementation.go
```diff
@@ -0,0 +1,241 @@
...
+		return false, nil
+	}
+
+	obj, ok := a[rwt.Identifier]
+	if !ok {
+		return true, api.ErrNotFound
+	}
+
+	if op == tagOperationCreate {
+		obj.Tags[rwt.Tag] = true
```
LittleFox94, line 233: I think Create returns a 422 error when the requested tag is already on the resource, but not sure if need to match that behavior - should at least be a documented deviation. add a comment

pkg/api/mock/mock_api_implementation_test.go
```diff
@@ -0,0 +1,226 @@
...
+		})
+
+		It("returns ErrNotFound when object to Destroy not existing", func() {
+			err := a.Destroy(context.TODO(), &vlanv1.VLAN{Identifier: "nonexistent-identifier"})
+			Expect(err).To(MatchError(api.ErrNotFound))
+		})
+
+	})
+
+	Context("Custom gomega matcher", Ordered, func() {
```
LittleFox94, line 130: any chance to extract some of those into example tests to have them in the docs?

LittleFox94 (changes requested):
Good work, but some things needs to be changed

pkg/api/mock/matcher/example_test.go
```diff
@@ -0,0 +1,92 @@
...
+		log.Fatalf("failed to create record: %s", err)
+	}
+
+	if err := a.Destroy(context.TODO(), &clouddnsv1.Record{Identifier: record.Identifier}); err != nil {
+		log.Fatalf("failed destroy record: %s", err)
+	}
+
+	ok := Expect(a.All()).To(
+		ContainElement(
+			Destroyed(1),
```
LittleFox94, line 57: please combine this with a check to make sure our Record was deleted, as this would be the most common usage of this

pkg/api/mock/matcher/example_test.go
```diff
@@ -0,0 +1,92 @@
...
+
+	a := mock.NewMockAPI()
+
+	if err := a.Create(context.TODO(), &clouddnsv1.Record{ZoneName: "example.com", Name: "webmail"}); err != nil {
+		log.Fatalf("failed to create DNS record: %s", err)
+	}
+
+	ok := Expect(a.Existing()).To(
+		ContainElement(
+			Object(&clouddnsv1.Record{ZoneName: "example.com"}, "ZoneName"),
```
LittleFox94, line 29: either add a comment like "we'd have to check if it's existing, too" or just the check for that

LittleFox94: ah scratch that, this is already done by using `a.Existing()` - maybe add a comment clarifying this though ^^

pkg/api/mock/matcher/example_test.go
```diff
@@ -0,0 +1,92 @@
...
+	if err := a.Create(context.TODO(), zone); err != nil {
+		log.Fatalf("failed to create DNS zone: %s", err)
+	}
+
+	if err := corev1.Tag(context.Background(), a, zone, "Tag-0", "Tag-1", "Tag-2"); err != nil {
+		log.Fatalf("failed to tag zone: %s", err)
+	}
+
+	ok := Expect(a.Existing()).To(
+		ContainElement(
```
LittleFox94, line 84: again please add a line checking it's our record that has this tags. Not part of this test, but good part of this example

LittleFox94 (changes requested):
All my "please change" comments aside, I tried using this in LBaaS reconciliation tests in CCM and I love the diff - great work so far!

(https://github.com/anexia-it/k8s-anexia-ccm/pull/101/files)

pkg/api/mock/mock_api_object.go
```diff
@@ -0,0 +1,69 @@
+package mock
+
+import (
+	"time"
+
+	"go.anx.io/go-anxcloud/pkg/api/types"
+)
+
+// APIObject wraps types.Object with mock specific metadata
+type APIObject struct {
```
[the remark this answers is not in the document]
marioreggiori: APIObject is used in the `matcher` sub-package. I gues we could make the struct private and only expose an interface?

pkg/api/mock/mock_api_iterator.go
marioreggiori: Most endpoints seem to have no limit when page size (limit) is set to zero. But ENGSUP-6081 say's we shouldn't use it. So I think it's okay to return an error.

pkg/api/mock/matcher_object_tags.go
marioreggiori: I've missed that part. I think we only need this in `FailureMessage` as `NegatedFailureMessage` is only displayed when $\text{searched tags} \subset \text{object tags}$ and it's negated.

pkg/api/mock/mock_api_object.go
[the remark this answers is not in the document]
marioreggiori: > I feel like there are quite some things that might not need to be exported, can you check everything exported again and think over that? Or tell me to do that? ^^

I couldn't find anything else that is exported but shouldn't. Do you have anything specific (besides APIObject) which might not need to be exported?

pkg/api/mock/matcher/example_test.go
```diff
@@ -0,0 +1,98 @@
...
+
+	if err := a.Create(context.TODO(), &record); err != nil {
+		log.Fatalf("failed to create record: %s", err)
+	}
+
+	if err := a.Destroy(context.TODO(), &clouddnsv1.Record{Identifier: record.Identifier}); err != nil {
+		log.Fatalf("failed destroy record: %s", err)
+	}
+
+	ok := Expect(a.All()).To(
```
LittleFox94, line 56: In a real Ginkgo/Gomega test we don't use the return value, maybe we should just print "Success" after this line - if it doesn't succeed, the matcher panics and this is recovered by Gomega

pkg/api/mock/matcher/example_test.go
```diff
@@ -0,0 +1,98 @@
...
+	defer GinkgoRecover()
+
+	a := mock.NewMockAPI()
+
+	if err := a.Create(context.TODO(), &clouddnsv1.Record{ZoneName: "example.com", Name: "webmail"}); err != nil {
+		log.Fatalf("failed to create DNS record: %s", err)
+	}
+
+	// a.Existing() retrieves all objects which aren't currently flagged as destroyed
+	ok := Expect(a.Existing()).To(
```
LittleFox94, line 28: In a real Ginkgo/Gomega test we don't use the return value, maybe we should just print "Success" after this line - if it doesn't succeed, the matcher panics and this is recovered by Gomega

pkg/api/mock/matcher/example_test.go
```diff
@@ -0,0 +1,98 @@
...
+
+	if err := a.Create(context.TODO(), &zone); err != nil {
+		log.Fatalf("failed to create DNS zone: %s", err)
+	}
+
+	if err := corev1.Tag(context.Background(), a, &zone, "Tag-0", "Tag-1", "Tag-2"); err != nil {
+		log.Fatalf("failed to tag zone: %s", err)
+	}
+
+	ok := Expect(a.Existing()).To(
```
LittleFox94, line 86: In a real Ginkgo/Gomega test we don't use the return value, maybe we should just print "Success" after this line - if it doesn't succeed, the matcher panics and this is recovered by Gomega

pkg/api/mock/matcher/example_test.go
```diff
@@ -0,0 +1,98 @@
...
+		log.Fatalf("failed destroy record: %s", err)
+	}
+
+	ok := Expect(a.All()).To(
+		ContainElement(
+			Destroyed(1),
+		),
+	)
+
+	Expect(a.Inspect(record.Identifier).Existing()).To(BeFalse())
```
LittleFox94, line 62: I guess this was added in response to this https://github.com/anexia-it/go-anxcloud/pull/139/files#r907387666

I added another comment to that thread to clarify what I was meaning

LittleFox94 (changes requested):

pkg/api/mock/matcher/matcher_object_tags.go
```diff
@@ -0,0 +1,52 @@
...
+
+func (m *objectTaggedWith) FailureMessage(actual interface{}) string {
+	tags, err := extractObjectTags(actual)
+	if err != nil {
+		return fmt.Sprintf("Failed to extract object tags: %s", err)
+	}
+
+	subFailureMessage := gomega.ContainElements(m.tags).FailureMessage(tags)
+
+	return fmt.Sprintf("Object not tagged with provided tags: %s", subFailureMessage)
```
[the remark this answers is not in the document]
marioreggiori: Not sure if we should call them `required` as the negated version:
```
      Object tagged with all required tags: Expected
          <[]string | len:2, cap:2>: ["tag-3", "tag-2"]
      not to contain elements
          <[]string | len:2, cap:2>: ["tag-2", "tag-3"]
```
could be missleading.

pkg/api/mock/matcher/matcher_object.go
```diff
@@ -0,0 +1,76 @@
...
+	if reflect.TypeOf(m.obj) != reflect.TypeOf(obj.Unwrap()) {
+		return nil, ErrObjectMatcherTypeMismatch
+	}
+
+	diff, err := compare.Compare(obj.Unwrap(), m.obj, m.matchFields...)
+	if err != nil {
+		return nil, err
+	}
+
+	return diff, nil
```
LittleFox94, line 75: We currently get an output like this:
`ACTUAL Object does not match EXPECTED Object: []compare.Difference{compare.Difference{Key:"Name", A:"test", B:"foo"}}`

But with a simple change, we can get a much nicer output like this:
```
ACTUAL Object does not match EXPECTED Object:
  Key "Name"
  Expected
      <string>: test
  to equal
      <string>: foo
```

```golang
	ret := make([]string, 0, len(diff))
	for _, d := range diff {
		ret = append(ret, fmt.Sprintf(
			"Key %q\n%s\n\n",
			d.Key,
			format.Message(d.A, "to equal", d.B),
		))
	}

	return ret, nil
```

(combined with returning a `[]string` and using `strings.Join(diff, "\n")` in `FailureMessage()`.

pkg/api/mock/matcher/matcher_test.go
```diff
@@ -0,0 +1,187 @@
...
+
+		It("can match objects destroyed n times", func() {
+			Expect(ContainElement(Destroyed(5)).Match(a.All())).To(BeFalse())
+			Expect(ContainElement(Destroyed(0)).Match(a.All())).To(BeTrue())
+
+			obj := testObject{}
+			a.FakeExisting(&obj)
+			err := a.Destroy(context.TODO(), &obj)
+			Expect(err).ToNot(HaveOccurred())
+			Expect(ContainElementNTimes(Destroyed(1), 1).Match(a.All())).To(BeTrue())
```
LittleFox94, line 101: can you add a test case for `Destroyed(2)`? Forgot why we have that as number, in which case we destroy the same Object multiple times.. ^^'

pkg/api/mock/mock_api_implementation.go
```diff
@@ -0,0 +1,289 @@
...
+		mao.tags[tag] = true
+	}
+
+	a.data[identifier] = &mao
+
+	return identifier
+}
+
+// Inspect retrieves an *APIObject or nil if not found
+func (a *mockAPI) Inspect(id string) *APIObject {
```
LittleFox94, line 284: Please call this `identifier` for clarity/convention.

pkg/api/mock/mock_api_implementation_test.go
```diff
@@ -0,0 +1,241 @@
...
+			Expect(a.Inspect(id).Unwrap()).To(Equal(&obj))
+		})
+
+		It("returns nil if id not in API", func() {
+			Expect(a.Inspect("not-in-api")).To(BeNil())
+		})
+	})
+
+	Context("Get operations", func() {
+		It("returns api.ErrUnidentifiedObject if object does not have an identifier set", func() {
```
LittleFox94, line 44: `types.ErrUnidentifiedObject`?

pkg/api/mock/mock_api_implementation_test.go
```diff
@@ -0,0 +1,241 @@
...
+			obj := testObject{}
+			err := a.Create(context.TODO(), &obj)
+			Expect(err).ToNot(HaveOccurred())
+			Expect(a.All().Unwrap()).To(ContainElement(&obj))
+		})
+
+		It("supports corev1.ResourceWithTag", func() {
+			id := a.FakeExisting(&testObject{})
+			err := a.Create(context.TODO(), &corev1.ResourceWithTag{Identifier: id, Tag: "tagname"})
+			Expect(err).ToNot(HaveOccurred())
```
LittleFox94, line 178: add a check for APIObject having the tag now

pkg/api/mock/mock_api_implementation_test.go
```diff
@@ -0,0 +1,241 @@
...
+		It("can destroy objects", func() {
+			id := a.FakeExisting(&testObject{})
+			err := a.Destroy(context.TODO(), &testObject{Identifier: id})
+			Expect(err).ToNot(HaveOccurred())
+		})
+
+		It("supports corev1.ResourceWithTag", func() {
+			id := a.FakeExisting(&testObject{}, "tagname")
+			err := a.Destroy(context.TODO(), &corev1.ResourceWithTag{Identifier: id, Tag: "tagname"})
+			Expect(err).ToNot(HaveOccurred())
```
LittleFox94, line 222: add a check for the `APIObject` now not containing the tag anymore

pkg/api/mock/utils_test.go
```diff
@@ -0,0 +1,101 @@
...
+			Expect(err).ToNot(HaveOccurred())
+			Expect(res).To(Equal(&test{String: "test", Int: 1}))
+
+			res, err = merge(&test{String: "test", Int: 1}, &test{})
+			Expect(err).ToNot(HaveOccurred())
+			Expect(res).To(Equal(&test{String: "test", Int: 1}))
+
+			res, err = merge(&test{String: "test", Int: 1}, &test{String: "test_changed", Int: 2})
+			Expect(err).ToNot(HaveOccurred())
+			Expect(res).To(Equal(&test{String: "test_changed", Int: 2}))
```
LittleFox94, line 35: this check is now practically the same first one, the second Object overriding all properties - change it to only updating one field and checking the other field is still the same after merge

pkg/api/mock/utils_test.go
```diff
@@ -0,0 +1,101 @@
...
+			Expect(err).ToNot(HaveOccurred())
+			Expect(res).To(Equal(&test{String: &s, Int: &i}))
+
+			res, err = merge(&test{String: &s, Int: &i}, &test{})
+			Expect(err).ToNot(HaveOccurred())
+			Expect(res).To(Equal(&test{String: &s, Int: &i}))
+
+			res, err = merge(&test{String: &s, Int: &i}, &test{String: &sChanged, Int: &iChanged})
+			Expect(err).ToNot(HaveOccurred())
+			Expect(res).To(Equal(&test{String: &sChanged, Int: &iChanged}))
```
LittleFox94, line 64: same as in `can merge primitives"

pkg/api/mock/utils_test.go
```diff
@@ -0,0 +1,101 @@
...
+				String string `json:"string,omitempty"`
+			}
+			type test struct {
+				Nested nested `json:"nested,omitempty"`
+			}
+
+			res, err = merge(&test{}, &test{Nested: nested{String: "test"}})
+			Expect(err).ToNot(HaveOccurred())
+			Expect(res).To(Equal(&test{Nested: nested{String: "test"}}))
+
```
LittleFox94, line 81: do we need some more test cases here, as in the others above?

LittleFox94 (changes requested):
This has a lot of nitpicks now, but I think worst is done :)

pkg/api/mock/matcher/matcher_test.go
marioreggiori: CloudDNS Zones have their name as identifier and can potentially be created & destroyed multiple times.

pkg/api/mock/internal/core_types.go
```diff
@@ -0,0 +1,30 @@
...
+	"go.anx.io/go-anxcloud/pkg/api/types"
+	clouddnsv1 "go.anx.io/go-anxcloud/pkg/apis/clouddns/v1"
+	corev1 "go.anx.io/go-anxcloud/pkg/apis/core/v1"
+	lbaasv1 "go.anx.io/go-anxcloud/pkg/apis/lbaas/v1"
+	vlanv1 "go.anx.io/go-anxcloud/pkg/apis/vlan/v1"
+)
+
+// ObjectCoreType maps types.Object to *corev1.Type
+// returns nil if no mapping was found
+// TODO: (SYSENG-1333) have generic client bindings register themselves with their Type Identifier so we can use that for other purposes, too
```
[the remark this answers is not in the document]
LittleFox94: yeah yeah, I requested that TODO :D

pkg/api/mock/matcher/matcher_object.go
LittleFox94: oh yeah, better do that in `FailureMessage`, we don't need to do that for `Match` or `NegatedFailureMessage` after all - good catch!

pkg/api/mock/utils_test.go
LittleFox94: I would have updated the string attribute, but that's not important ^^

LittleFox94 (changes requested):
nearly there! \o/

LittleFox94 (approved):
lgtm!

even though it's still a draft? ;)

PR #148, opened 2022-06-17
Add generic vSphere Template bindings and FindNamedTemplate helper

marioreggiori:
### Description

Add generic vSphere Template bindings and FindNamedTemplate helper to mitigate deleted VM template (ENGSUP-5913) issues.

### Checklist

### Community Note

### DIFF
--- a/pkg/apis/vsphere/v1/template_genclient.go
+++ b/pkg/apis/vsphere/v1/template_genclient.go
@@ -1,0 +1,102 @@
+package v1
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"io"
+	"net/url"
+	"path"
+
+	"go.anx.io/go-anxcloud/pkg/api"
+	"go.anx.io/go-anxcloud/pkg/api/types"
+)
+
+var (
+	// ErrTemplateOperationNotSupported is returned if operations other than Get and List are performed
+	ErrTemplateOperationNotSupported = fmt.Errorf("%w: Template only supports Get and List operations", api.ErrOperationNotSupported)
+)
+
+// EndpointURL returns the URL where to retrieve objects of type Template (only Get and List operations supported)
+func (t *Template) EndpointURL(ctx context.Context) (*url.URL, error) {
+	_, err := t.operationFromContext(ctx)
+	if err != nil {
+		return nil, err
+	}
+
+	return url.ParseRequestURI(
+		fmt.Sprintf("/api/vsphere/v1/provisioning/templates.json/%s/%s", t.Location.Identifier, t.Type),
+	)
+}
+
+// HasPagination disables pagination for Template API (not supported by engine)
+func (t *Template) HasPagination(ctx context.Context) (bool, error) {
+	return false, nil
+}
+
+// FilterRequestURL removes the Identifier from URL on Get operations (template needs to be parsed from list response)
+func (t *Template) FilterRequestURL(ctx context.Context, url *url.URL) (*url.URL, error) {
+	op, err := t.operationFromContext(ctx)
+	if err != nil {
+		return nil, err
+	}
+
+	if op == types.OperationGet {
+		url.Path = path.Dir(url.Path)
+	}
+
+	q := url.Query()
+	q.Set("page", "1")
+	q.Set("limit", "1000")
+	url.RawQuery = q.Encode()
+
+	return url, nil
+}
+
+// DecodeAPIResponse is used to filter a single template on Get operations
+func (t *Template) DecodeAPIResponse(ctx context.Context, data io.Reader) error {
+	op, err := t.operationFromContext(ctx)
+	if err != nil {
+		return err
+	}
+
+	if op == types.OperationGet {
+		tpl, err := t.extractTemplateByID(data)
+		if err != nil {
+			return err
+		}
+		*t = *tpl
+		return nil
+	}
+
+	return json.NewDecoder(data).Decode(t)
+}
+
+func (t *Template) extractTemplateByID(data io.Reader) (*Template, error) {
+	var templates []*Template
+	err := json.NewDecoder(data).Decode(&templates)
+	if err != nil {
+		return nil, err
+	}
+
+	for _, template := range templates {
+		if template.Identifier == t.Identifier {
+			return template, nil
+		}
+	}
+
+	return nil, fmt.Errorf("Template with given Identifier not found in response: %w", api.ErrNotFound)
+}
+
+func (t *Template) operationFromContext(ctx context.Context) (types.Operation, error) {
+	op, err := types.OperationFromContext(ctx)
+	if err != nil {
+		return "", err
+	}
+
+	if op != types.OperationGet && op != types.OperationList {
+		return "", ErrTemplateOperationNotSupported
+	}
+
+	return op, nil
+}
--- a/pkg/apis/vsphere/v1/template_helpers.go
+++ b/pkg/apis/vsphere/v1/template_helpers.go
@@ -1,0 +1,75 @@
+package v1
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/go-logr/logr"
+	"go.anx.io/go-anxcloud/pkg/api"
+	"go.anx.io/go-anxcloud/pkg/api/types"
+	corev1 "go.anx.io/go-anxcloud/pkg/apis/core/v1"
+)
+
+var (
+	// ErrTemplateNotFound is returned when the named template was not found at a given location
+	ErrTemplateNotFound = fmt.Errorf("%w: named template was not found at specified location", api.ErrNotFound)
+)
+
+const (
+	// LatestTemplateBuild is used to find the template with the highest build number
+	LatestTemplateBuild = "latest"
+)
+
+// FindNamedTemplate retrieves a template by name and build at a specified location.
+// Empty and LatestTemplateBuild build identifier will yield the highest available build.
+// It returns ErrTemplateNotFound if no matching template was found.
+func FindNamedTemplate(ctx context.Context, a api.API, name, build string, location corev1.Location) (*Template, error) {
+	var match *Template
+	buildNo := -1
+	latest := build == "" || build == LatestTemplateBuild
+
+	ctx, cancel := context.WithCancel(ctx)
+	defer cancel()
+
+	var channel types.ObjectChannel
+
+	err := a.List(ctx, &Template{Type: TypeTemplate, Location: location}, api.ObjectChannel(&channel))
+	if err != nil {
+		return nil, fmt.Errorf("error listing templates: %w", err)
+	}
+
+	for res := range channel {
+		var template Template
+		err := res(&template)
+		if err != nil {
+			return nil, fmt.Errorf("error retrieving template: %w", err)
+		}
+
+		if template.Name != name {
+			continue
+		}
+
+		if latest {
+			currentTemplateBuildNo, err := template.BuildNumber()
+			if err != nil {
+				logr.FromContextOrDiscard(ctx).Info("couldn't parse build %q of template %q from location %q", template.Build, template.Identifier, location.Identifier)
+				continue
+			}
+
+			if latest && (match == nil || currentTemplateBuildNo > buildNo) {
+				match = &template
+				buildNo = currentTemplateBuildNo
+			}
+		} else if template.Build == build {
+			match = &template
+			break
+		}
+
+	}
+
+	if match == nil {
+		return nil, fmt.Errorf("%w (name: %q, build: %q, location: %q)", ErrTemplateNotFound, name, build, location.Identifier)
+	}
+
+	return match, nil
+}
--- a/pkg/apis/vsphere/v1/template_test.go
+++ b/pkg/apis/vsphere/v1/template_test.go
@@ -1,0 +1,112 @@
+package v1
+
+import (
+	"context"
+	"fmt"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	"github.com/onsi/gomega/ghttp"
+	"go.anx.io/go-anxcloud/pkg/api"
+	"go.anx.io/go-anxcloud/pkg/api/types"
+	corev1 "go.anx.io/go-anxcloud/pkg/apis/core/v1"
+	"go.anx.io/go-anxcloud/pkg/client"
+)
+
+var _ = Describe("Template API bindings", func() {
+	Context("BuildNumber method", func() {
+		It("can parse valid Build identifier", func() {
+			t := Template{Build: "b04"}
+			Expect(t.BuildNumber()).To(Equal(4))
+		})
+
+		DescribeTable("invalid build identifier", func(build string) {
+			t := Template{Build: build}
+			_, err := t.BuildNumber()
+			Expect(err).To(MatchError(ErrFailedToParseTemplateBuildNumber))
+		},
+			Entry("with empty build string", ""),
+			Entry("without build digits", "b"),
+			Entry("with unknown build prefix", "c123"),
+			Entry("with characters between build digits", "b1c23"),
+		)
+	})
+
+	Context("with mocked server", func() {
+		var a api.API
+		location := corev1.Location{Identifier: "mock-location-id"}
+
+		BeforeEach(func() {
+			srv := ghttp.NewServer()
+			srv.AppendHandlers(ghttp.CombineHandlers(
+				ghttp.VerifyRequest("GET", fmt.Sprintf("/api/vsphere/v1/provisioning/templates.json/%s/templates", location.Identifier), "page=1&limit=1000"),
+				ghttp.RespondWithJSONEncoded(200, mockedTemplateList()),
+			))
+
+			var err error
+			a, err = api.NewAPI(api.WithClientOptions(
+				client.BaseURL(srv.URL()),
+				client.IgnoreMissingToken(),
+			))
+			Expect(err).ToNot(HaveOccurred())
+		})
+
+		It("can List templates", func() {
+			var channel types.ObjectChannel
+			err := a.List(context.TODO(), &Template{Type: TypeTemplate, Location: location}, api.ObjectChannel(&channel))
+			Expect(err).ToNot(HaveOccurred())
+
+			templateCount := 0
+			for res := range channel {
+				t := Template{}
+				err := res(&t)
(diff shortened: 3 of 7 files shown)


LittleFox94 (changes requested):

pkg/apis/vsphere/v1/template_types.go
```diff
@@ -0,0 +1,13 @@
...
+// anxcloud:object:hooks=PaginationSupportHook
+
+// Template represents a vSphere template used for vm provisioning
+type Template struct {
+	Identifier string `json:"id" anxcloud:"identifier"`
+	Name       string `json:"name"`
+	Bit        string `json:"bit"`
+	Build      string `json:"build"`
+	Location   string
+	Type       string
```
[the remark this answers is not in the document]
marioreggiori: The constant `Template` would clash with the `Template` API struct. The engine supports "templates" and "from_scratch" as valid inputs. Maybe `Templates` instead of `Template`?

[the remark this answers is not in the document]
LittleFox94: oh right ^^'

what about `TypeTemplate` and `TypeFromScratch`?

pkg/apis/vsphere/v1/template_genclient.go
```diff
@@ -0,0 +1,31 @@
...
+import (
+	"context"
+	"fmt"
+	"net/url"
+
+	"go.anx.io/go-anxcloud/pkg/api"
+	"go.anx.io/go-anxcloud/pkg/api/types"
+)
+
+// EndpointURL returns the URL where to retrieve objects of type Template (only list operations supported atm).
```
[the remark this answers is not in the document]
marioreggiori: Added support

pkg/apis/vsphere/v1/template_helpers.go
```diff
@@ -0,0 +1,69 @@
...
+	"fmt"
+	"strconv"
+
+	"go.anx.io/go-anxcloud/pkg/api"
+	"go.anx.io/go-anxcloud/pkg/api/types"
+)
+
+var (
+	// ErrTemplateNotFound is returned when the named template was not found at a given location
+	ErrTemplateNotFound = errors.New("named template was not found at specified location")
```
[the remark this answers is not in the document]
marioreggiori: Makes sense :+1:

pkg/apis/vsphere/v1/template_genclient.go
```diff
@@ -0,0 +1,97 @@
...
+	for _, template := range templates {
+		if template.Identifier == t.Identifier {
+			return template, nil
+		}
+	}
+
+	return nil, ErrTemplateNotFound
+}
+
+func (t *Template) operationFromContext(ctx context.Context) (types.Operation, error) {
```
LittleFox94, line 86: Huh why the extra method for that? Would still only be 3 returns in `EndpointURL`, right?

pkg/apis/vsphere/v1/template_types.go
```diff
@@ -0,0 +1,27 @@
...
+
+// TemplateType specifies the type of template
+type TemplateType string
+
+const (
+	// TypeTemplate is used for prebuilt templates
+	TypeTemplate = TemplateType("templates")
+	// TypeFromScratch is used for custom templates
+	TypeFromScratch = TemplateType("from_scratch")
+)
```
LittleFox94, line 27: This way the constants show up below the type in the documentation:

```go
const (
	// TypeTemplate is used for prebuilt templates
	TypeTemplate TemplateType = "templates"

	// TypeFromScratch is used for custom templates
	TypeFromScratch TemplateType = "from_scratch"
)
```

pkg/apis/vsphere/v1/template_genclient.go
```diff
@@ -0,0 +1,97 @@
...
+	"net/url"
+	"path"
+
+	"go.anx.io/go-anxcloud/pkg/api"
+	"go.anx.io/go-anxcloud/pkg/api/types"
+)
+
+var (
+	// ErrTemplateOperationNotSupported is returned if operations other than Get and List are performed
+	ErrTemplateOperationNotSupported = fmt.Errorf("%w: Template only support Get and List operations", api.ErrOperationNotSupported)
```
LittleFox94, line 17: supports*

pkg/apis/vsphere/v1/template_genclient.go
```diff
@@ -0,0 +1,97 @@
...
+)
+
+// EndpointURL returns the URL where to retrieve objects of type Template (only Get and List operations supported atm).
```
LittleFox94, line 20: strip the "atm" please

pkg/apis/vsphere/v1/template_genclient.go
```diff
@@ -0,0 +1,97 @@
...
+func (t *Template) EndpointURL(ctx context.Context) (*url.URL, error) {
+	_, err := t.operationFromContext(ctx)
+	if err != nil {
+		return nil, err
+	}
+
+	return url.ParseRequestURI(
+		fmt.Sprintf("/api/vsphere/v1/provisioning/templates.json/%s/%s?page=1&limit=1000", t.Location.Identifier, t.Type),
```
LittleFox94, line 28: Maybe we should return the URL without query parameters and add them in `FilterRequestURL`

pkg/apis/vsphere/v1/template_genclient.go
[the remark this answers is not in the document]
LittleFox94: I didn't realize there is no API for retrieving a template by its identifier ^^'

But now the code is here anyway x)

pkg/apis/vsphere/v1/template_genclient.go
```diff
@@ -0,0 +1,97 @@
...
+		return nil, err
+	}
+
+	for _, template := range templates {
+		if template.Identifier == t.Identifier {
+			return template, nil
+		}
+	}
+
+	return nil, ErrTemplateNotFound
```
LittleFox94, line 83: `ErrTemplateNotFound` is for "named template not found" but this case here isn't about named templates. Probably just return a `pkg/api.ErrNotFound` here, maybe wrap with "Template with given Identifier not found in response".

pkg/apis/vsphere/v1/template_helpers.go
```diff
@@ -0,0 +1,73 @@
...
+)
+
+// FindNamedTemplate retrieves a template by name and build at a specified location.
+// Empty and LatestTemplateBuild build identifier will yield the highest available build.
+// It returns ErrTemplateNotFound if no matching template was found.
+func FindNamedTemplate(ctx context.Context, a api.API, name, build string, location corev1.Location) (*Template, error) {
+	var (
+		match   *Template
+		buildNo = -1
+		latest  = build == "" || build == LatestTemplateBuild
```
LittleFox94, line 30: for those two you don't need `var` when just using `:=` and then it's only using var for `match`, so please change to
```go
var match *Template
buildNo := -1
// ...
```

pkg/apis/vsphere/v1/template_helpers.go
```diff
@@ -0,0 +1,73 @@
...
+	)
+
+	ctx, cancel := context.WithCancel(ctx)
+	defer cancel()
+
+	var channel types.ObjectChannel
+
+	err := a.List(ctx, &Template{Type: TypeTemplate, Location: location}, api.ObjectChannel(&channel))
+	if err != nil {
+		return nil, err
```
LittleFox94, line 40: wrap with an "error listing templates" or something

pkg/apis/vsphere/v1/template_helpers.go
```diff
@@ -0,0 +1,73 @@
...
+	}
+
+	for res := range channel {
+		var template Template
+		err := res(&template)
+		if err != nil {
+			return nil, err
```
LittleFox94, line 47: wrap with an "error retrieving template" or something

pkg/apis/vsphere/v1/template_helpers.go
```diff
@@ -0,0 +1,73 @@
...
+		}
+
+		if template.Name != name {
+			continue
+		}
+
+		if latest {
+			currentTemplateBuildNo, _ := strconv.Atoi(template.Build[1:])
```
LittleFox94, line 55: please check if `strconv.Atoi` was successful

LittleFox94: Maybe give `Template` a method `BuildNumber() int`, you could use this here and users of go-anxcloud could easily reuse the same parsing logic for something different.

pkg/apis/vsphere/v1/template_test.go
```diff
@@ -0,0 +1,94 @@
...
+			Expect(tpl.Name).To(Equal("Flatcar Linux Stable"))
+		})
+
+		DescribeTable("find named template", func(name, build, expectedID string) {
+			template, err := FindNamedTemplate(context.TODO(), a, name, build, corev1.Location{Identifier: locationID})
+			Expect(err).ToNot(HaveOccurred())
+			Expect(template.Identifier).To(Equal(expectedID))
+		},
+			Entry("latest with empty build", "Flatcar Linux Stable", "", "26a47eee-dc9a-4eea-b67a-8fb1baa2fcc0"),
+			Entry("latest with latest build", "Flatcar Linux Stable", "latest", "26a47eee-dc9a-4eea-b67a-8fb1baa2fcc0"),
```
LittleFox94, line 64: please use `LatestTemplateBuild` constant instead of `"latest"`

pkg/apis/vsphere/v1/template_test.go
```diff
@@ -0,0 +1,94 @@
...
+
+func mockedTemplateList() []Template {
+	return []Template{
+		{Identifier: "e9325be9-25b9-468e-851e-56b5c0367e5a", Name: "Ubuntu 21.04", Build: "b72"},
+		{Identifier: "b21b8b77-30e3-478a-9b6d-1f61d29e9f9a", Name: "Flatcar Linux Stable", Build: "b73"},
+		{Identifier: "ec547552-d453-42e6-987d-51abe703c439", Name: "Debian 11", Build: "b18"},
+		{Identifier: "26a47eee-dc9a-4eea-b67a-8fb1baa2fcc0", Name: "Flatcar Linux Stable", Build: "b74"},
+		{Identifier: "cb16dc94-ec55-4e9a-a1a3-b76a91bbe274", Name: "Windows 2022", Build: "b06"},
+		{Identifier: "fc3a63c6-6f4e-4193-b368-ebe9e08b4302", Name: "Debian 10", Build: "b80"},
+		{Identifier: "844ac596-5f62-4ed2-936e-b99ffe0d4f88", Name: "Flatcar Linux Stable", Build: "b72"},
```
LittleFox94, line 89: Add an entry for your unchecked `strconv.Atoi`, it should just ignore this one
```go
{Identifier: $randomUUID, Name: "Flatcar Linux Stable", Build: "invalid-build-id"},
```

pkg/apis/vsphere/v1/template_test.go
```diff
@@ -0,0 +1,94 @@
...
+		{Identifier: "c3d4f0a6-978a-49fb-a952-7361bf531e4f", Name: "Debian 9", Build: "b92"},
+		{Identifier: "086c5f99-1be6-46ec-8374-cdc23cedd6a4", Name: "Windows 2022", Build: "b12"},
+		{Identifier: "9d863fd9-d0d3-4959-b226-e73192f3e43d", Name: "Debian 11", Build: "possibly-valid-build-id"},
```
LittleFox94, line 92: .. instead of adding the invalid-build-id flatcar, your "latest" tests could go for "Debian 11" instead

pkg/apis/vsphere/v1/template_test.go
```diff
@@ -0,0 +1,94 @@
...
+
+		It("can List templates", func() {
+			var channel types.ObjectChannel
+			err := a.List(context.TODO(), &Template{Type: TypeTemplate, Location: corev1.Location{Identifier: locationID}}, api.ObjectChannel(&channel))
+			Expect(err).ToNot(HaveOccurred())
+
+			templateCount := 0
+			template := &Template{}
+			for res := range channel {
+				err := res(template)
```
LittleFox94, line 44:
```go
for res := range channel {
    t := Template{}
    err := res(&t)
```

at the very least, do not declare `template` as `*Template`, to match common usage

pkg/apis/vsphere/v1/template_test.go
```diff
@@ -0,0 +1,94 @@
...
+			a, err = api.NewAPI(api.WithClientOptions(
+				client.BaseURL(srv.URL()),
+				client.IgnoreMissingToken(),
+			))
+			Expect(err).ToNot(HaveOccurred())
+		})
+
+		It("can List templates", func() {
+			var channel types.ObjectChannel
+			err := a.List(context.TODO(), &Template{Type: TypeTemplate, Location: corev1.Location{Identifier: locationID}}, api.ObjectChannel(&channel))
```
LittleFox94, line 38: you are creating a lot of the same `corev1.Location` objects - maybe just extract this as a variable all tests can use?

pkg/apis/vsphere/v1/template_genclient.go
marioreggiori: To de-duplicate `ErrTemplateOperationNotSupported` handling

pkg/apis/vsphere/v1/template_helpers.go
marioreggiori: Should the `BuildNumber` method also return an error when Atoi wasn't successful or default to zero-val?

LittleFox94: hmm... it should either try to log a warning (`logr.FromContextOrDiscard` or something) and skip the template or error out - defaulting to zero-val isn't good

Either way, it should only do that when "latest" template is requested.

Let's return an error

pkg/apis/vsphere/v1/template_genclient.go
```diff
@@ -0,0 +1,101 @@
...
+		return nil, err
+	}
+
+	if op == types.OperationGet {
+		url.Path = path.Dir(url.Path)
+	}
+
+	q := url.Query()
+	q.Set("page", "1")
+	q.Set("limit", "1000")
```
LittleFox94, line 50: is the query on the URL actually updated with this? I seem to remember it's not and you have to do something like this:
```go
url.RawQuery = q.Encde()
```

pkg/apis/vsphere/v1/template_helpers.go
```diff
@@ -0,0 +1,73 @@
...
+		}
+
+		if template.Name != name {
+			continue
+		}
+
+		if latest {
+			currentTemplateBuildNo, err := template.BuildNumber()
+			if err != nil {
+				continue
```
LittleFox94, line 54: I think erroring out is what we want here, as we cannot know if we return the best possible template in this case

pkg/apis/vsphere/v1/template_types.go
```diff
@@ -0,0 +1,51 @@
...
+	"errors"
+	"fmt"
+	"strconv"
+
+	corev1 "go.anx.io/go-anxcloud/pkg/apis/core/v1"
+)
+
+var (
+	// ErrFailedToParseTemplateBuildNumber is returned when the template build couldn't be converted to an int
+	ErrFailedToParseTemplateBuildNumber = errors.New("failed to parse template build-number")
```
LittleFox94, line 13: `build number`*

pkg/apis/vsphere/v1/template_test.go
```diff
@@ -0,0 +1,112 @@
...
+			Expect(err).ToNot(HaveOccurred())
+			Expect(tpl.Name).To(Equal("Flatcar Linux Stable"))
+		})
+
+		DescribeTable("find named template", func(name, build, expectedID string) {
+			template, err := FindNamedTemplate(context.TODO(), a, name, build, location)
+			Expect(err).ToNot(HaveOccurred())
+			Expect(template.Identifier).To(Equal(expectedID))
+		},
+			Entry("latest with empty build", "Debian 11", "", "ec547552-d453-42e6-987d-51abe703c439"),
```
LittleFox94, line 81: with `BuildNumber()` returning an error and `FindNamedTemplate()` also returning that error, this will, of course, not succeed anymore and needs to be adapted.

LittleFox94 (changes requested):
We are getting there, only small things now - good job :)

pkg/apis/vsphere/v1/template_helpers.go
LittleFox94: Didn't see such build numbers yet and doubt they exist, but maybe .. can you at least log a warning? `ctx` should always have a logger attached (`logr.FromContextOrDiscard()`)

pkg/apis/vsphere/v1/template_test.go
marioreggiori: This `Entry` does not create a `Template` on which `BuildNumber` is called. The empty build string is an alias for `LatestTemplateBuild` in `FindNamedTemplate`.

pkg/apis/vsphere/v1/template_test.go
marioreggiori: Is this resolved?

LittleFox94: yep, somehow missed this

pkg/apis/vsphere/v1/template_test.go
LittleFox94: uhm.. `BuildNumber()` is called whenever we want the latest template, so in this case it is called?

But since we decided to log a warning in `FindNamedTemplate` and not error out, this does not produce an error and does not have to be changed

marioreggiori: `BuildNumber` is only called on the templates we receive from the API. The empty build string from this `Entry` argument is never (tried to be) converted to an `int`.

pkg/apis/vsphere/v1/template_genclient.go
LittleFox94: did you check if this is needed?

marioreggiori: Yes, it's needed. `url.String()` did not contain the query params without.

LittleFox94:
Do you think it's feasible to add some more tests for all the currently-untested error conditions?

LittleFox94 (approved):

PR #150, opened 2022-07-11
Allow operation dependent object identification

marioreggiori:
### Description

Some Objects have multiple identifying fields. The corev1.Location for example can be identified via the common Identifier or alternatively via Code (e.g. "ANX04"). Primary identifying fields are tagged with `anxcloud:"identifier"`. This PR includes a `(x Object) GetIdentifier(context.Context) (string, error)` method generator which returns the primary Identifier. If there is already a `GetIdentifier` method for an Object the code generator will skip the object.

### Checklist

### Community Note


pkg/api/api_implementation.go
```diff
@@ -257,9 +257,11 @@ func (a defaultAPI) makeRequest(ctx context.Context, obj types.Object, body inte
 
 	// We do this right on top because this checks if the Object has a correct type which is more strictly defined than just the interface.
 	// In a perfect world this would be a compile-time check.
```
LittleFox94, line 259: This comment is out-of-date now. The original purpose of having "this right on top" was to check if the passed Object actually is an Object, which also means it has an Identifier attribute.

With the new Object.GetIdentifier() method we have a compile-time check for this.

What we actually check here is if the Identifier is set when required and give the Object a way to deny some operations with unsuitable Identifiers being set - please update the comment accordingly :)

Makefile
```diff
@@ -15,6 +15,8 @@ build: fmtcheck go-lint
 generate: tools
 	# generate object tests
 	tools/tools object-generator --mode tests --in ./pkg/... --out xxgenerated_object_test.go
+	# generate GetIdentifier methods for objects
+	tools/tools object-identifier
```
LittleFox94, line 19: if you look at the command above, you can see the `--mode tests` - this was intended to allow the code generator to run for "objects" in different modes, like "tests" and "runtime". The new generator you built is a part of a new "runtime"-mode "object-generator".

It should support the same `--in` and `--out` as the current object-generator in tests-mode.

pkg/api/mock/common_test.go
```diff
@@ -12,15 +14,20 @@ type testObject struct {
...
+func (o *testObject2) GetIdentifier(context.Context) (string, error)     { return o.Identifier, nil }
 
 type testObjectWithoutIdentifier struct{}
 
 func (o *testObjectWithoutIdentifier) EndpointURL(ctx context.Context) (*url.URL, error) {
 	return nil, nil
 }
+func (o *testObjectWithoutIdentifier) GetIdentifier(context.Context) (string, error) {
+	return "", types.ErrObjectWithoutIdentifier
+}
```
LittleFox94, line 33: Is this test object still relevant with now having a compile-time check if it is a full Object?

pkg/api/mock/matcher/common_test.go
```diff
@@ -12,13 +14,18 @@ type testObject struct {
...
 }
 
 func (o *testObject2) EndpointURL(ctx context.Context) (*url.URL, error) { return nil, nil }
+func (o *testObject2) GetIdentifier(context.Context) (string, error)     { return o.Identifier, nil }
 
 type testObjectNotStruct string
 
 func (o *testObjectNotStruct) EndpointURL(ctx context.Context) (*url.URL, error) { return nil, nil }
+func (o *testObjectNotStruct) GetIdentifier(context.Context) (string, error) {
+	return "", types.ErrObjectWithoutIdentifier
```
LittleFox94, line 30: the purpose of this test object is to check how we handle non-struct Objects - please return the string itself as identifier.

Maybe we have to add a "is struct" check somewhere in generic client or strip the requirement of Objects being structs - which was only in place to find the Identifier attribute in them. IIRC there is outside code relying on them being structs though.. not sure what to do, yet.

pkg/apis/core/v1/location_types.go
```diff
@@ -12,3 +18,21 @@ type Location struct {
...
+	if err != nil {
+		return "", nil
+	}
+
+	if op == types.OperationGet {
+		return l.Code, nil
+	}
+
+	return "", nil
+}
```
LittleFox94, line 38: this belongs in `location_genclient.go`

pkg/api/pagination_test.go
```diff
@@ -19,6 +19,10 @@ func (p paginationTestObject) EndpointURL(ctx context.Context) (*url.URL, error)
 	return nil, ErrOperationNotSupported
 }
 
+func (paginationTestObject) GetIdentifier(context.Context) (string, error) {
+	return "", ErrObjectWithoutIdentifier
```
LittleFox94, line 23: not that it's actually important here, but return the Message as Identifier instead of an error

Makefile
LittleFox94: You can integrate your new code into the existing object_generator code, reducing a lot of duplicate code. Directory traversal, "code is generated!"-header and co is all already done.

Ping me if you need pointers how to do that.

LittleFox94 (changes requested):

Makefile
```diff
@@ -15,6 +15,8 @@ build: fmtcheck go-lint
 generate: tools
 	# generate object tests
 	tools/tools object-generator --mode tests --in ./pkg/... --out xxgenerated_object_test.go
+	# generate GetIdentifier methods for objects
+	tools/tools object-generator --mode object-generator --in ./pkg/... --out xxgenerated_object_identifier.go
```
LittleFox94, line 19: object-generator in mode object-generator sounds strange ... why not the suggested "runtime"?

LittleFox94, line 19: please use `xxgenerated_object.go` as output file, other "runtime" code being generated for objects would also go in there.

pkg/api/mock/common_test.go
LittleFox94: nice catch making it a check for failing `GetIdentifier()`!

pkg/api/object_example_test.go
```diff
@@ -42,13 +42,26 @@ func (o *ExampleObject) EndpointURL(ctx context.Context) (*url.URL, error) {
 	return url.Parse("/example/v1")
 }
 
+// GetIdentifier returns the identifier of an object
+// The Makefiles `generate` target creates this method for objects that have a anxcloud:"identifier" tagged field
+// Objects with multiple identifying fields can have this method overwritten to return an operation dependent identifier
+// The provided context might not always have the operation set. In this case the method must return the primary identifier
```
LittleFox94, line 48: Those sentences all lack a `.` at the end - which is fine for me when it's the only sentence in a comment, but not in this case.

pkg/api/object_example_test.go
```diff
@@ -42,13 +42,26 @@ func (o *ExampleObject) EndpointURL(ctx context.Context) (*url.URL, error) {
...
+func (o *ExampleObject) GetIdentifier(ctx context.Context) (string, error) {
```
LittleFox94, line 49: Maybe we should extend this example for showing what other identifier might be possible, how to retrieve the operation and what we mean with "primary identifier"?

tools/object_generator.go
```diff
@@ -68,6 +71,10 @@ func (gen *ObjectGenerator) run() {
 	}
 
 	for _, s := range sources {
+		if gen.mode == "object-generator" && s == gen.outfile {
+			continue
+		}
+
```
LittleFox94: interesting .. so we skip parsing the output file, but only for our runtime mode? In which corner case did you run, is this really specific to this mode or should this be done for all modes?

tools/object_generator.go
```diff
@@ -233,6 +250,7 @@ func (gen *ObjectGenerator) processComment(comment string, decl *ast.GenDecl, fi
 				Line:     line,
 			})
 		}
+
```
LittleFox94, line 253: is this here for a reason?

pkg/apis/clouddns/v1/xxgenerated_object.go
```diff
@@ -0,0 +1,15 @@
+package v1
```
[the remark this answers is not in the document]
LittleFox94: configure code climate to ignore files matching xxgenerated_*.go

this should help: https://docs.codeclimate.com/docs/advanced-configuration#exclude-patterns

tools/object_generator.go
```diff
@@ -125,6 +137,11 @@ import (
 	"go.anx.io/go-anxcloud/pkg/api/types"
 )
 `)
+	} else if gen.mode == "runtime" {
```
LittleFox94, line 140: the `if gen.mode == "docs"` above is a leftover from a idea I had when starting this code generator .. but that comment it adds is useful for all generated go code, so please add it for modes `tests` and `runtime`.

tools/object_identifier.go
```diff
@@ -0,0 +1,89 @@
...
+type identifiableObject struct {
+	ObjectName       string
+	IdentifyingField string
+}
+
+func (gen *ObjectGenerator) generateGetIdentifier() {
+	const templates = `
+{{range .IdentifiedObjects }}
+// GetIdentifier returns the primary identifier of a {{.ObjectName}} object
+func (x *{{.ObjectName}}) GetIdentifier(ctx context.Context) (string, error) {
```
LittleFox94, line 21: small nitpick: change the receivers name from `x` to `o`, which is the common single char name for objects

pkg/api/object_example_test.go
LittleFox94: Also, these sentences all lack a ending `.`, too.

pkg/api/mock/matcher/matcher_test.go
```diff
@@ -146,13 +145,6 @@ var _ = Describe("custom gomega matcher", func() {
 			Expect(a.All()).To(ContainElementNTimes(Object(&testObject2{}), 1))
 		})
 
-		It("returns error when match is called with uncomparable type", func() {
```
LittleFox94, line 149: I think we can still have cases of `interface Object`-valid Objects that are incompatible with `pkg/utils/object/compare` and still need some checking for handling them correctly - so this case still makes sense I think

LittleFox94 (changes requested):

tools/object_generator.go
marioreggiori: Without this check, all objects would be marked as already identified in the next run.

LittleFox94: ah this makes sense, of course - thanks for clarifying :)

Still, I think we should exclude the output file from parsing for all modes, or do you see any case where we want to include things we generated in a previous run?

.codeclimate.yml
```diff
@@ -0,0 +1,5 @@
+exclude_patterns:
+  - "**/*_test.go"
+  - "**/test/"
+  - "**/tests/"
```
LittleFox94, line 4: not sure about those entries

pkg/api/object_example_test.go
```diff
@@ -42,13 +43,36 @@ func (o *ExampleObject) EndpointURL(ctx context.Context) (*url.URL, error) {
...
+//
+// For example: `corev1.Location` has the (primary) `Identifier` attribute, which is required by many API endpoints.
+// Additionally locations are identified by the human readable `Code` attribute (e.g. ANX04).
+//
+// Handling of secondary identifiers usually requires additional logic in the objects EndpointURL method.
+func (o *ExampleObject) GetIdentifier(ctx context.Context) (string, error) {
+	op, err := types.OperationFromContext(ctx)
+	if err == nil && op == types.OperationGet {
+		return o.SecondaryIdentifier, nil
+	}
```
LittleFox94, line 59: shouldn't this also check if `o.Identifier` is set and if so, return that and only otherwise look at secondary identifiers?

pkg/api/object_example_test.go
LittleFox94: ... did I really comment that twice? oh :deer:

LittleFox94 (changes requested):
Only small things left, good job!

.codeclimate.yml
marioreggiori: Without `**/test/` code climate was complaining about multiple files not related to this PR.

LittleFox94: And it's included in the default config, ok.

pkg/api/object_example_test.go
LittleFox94: you made this really nice for corev1.Location, checking if Identifier is set, if so returning it and only otherwise go to the secondary identifier logic. I'd love this example to look as great as that.

Also, please add the custom EndpointURL logic required for this to this example, too.

pkg/api/object_example_test.go
```diff
@@ -39,7 +40,43 @@ type ExampleObject struct {
 // identifier is appended by default for Get, Update and Destroy operations. You can implement the interface
 // types.RequestFilterHook to have full control over the requests done for your object.
 func (o *ExampleObject) EndpointURL(ctx context.Context) (*url.URL, error) {
-	return url.Parse("/example/v1")
+	op, err := types.OperationFromContext(ctx)
+	if err != nil {
+		return nil, err
```
LittleFox94, line 45: we should encourage wrapping errors before they are returned

pkg/api/object_example_test.go
```diff
@@ -39,7 +40,43 @@ type ExampleObject struct {
...
+// Handling of secondary identifiers usually requires additional logic in the objects EndpointURL method.
+func (o *ExampleObject) GetIdentifier(ctx context.Context) (string, error) {
+	if o.Identifier != "" {
+		return o.Identifier, nil
+	}
+
+	op, err := types.OperationFromContext(ctx)
+	if err != nil {
+		return "", nil
+	}
```
LittleFox94, line 73: first I stumbled over the `return "", nil` in error case, but then I remembered this is just the primary identifiers right now. We should make this more clear, but I think we can even improve the code some more:

```go
op, err := types.OperationFromContext(ctx)
if o.Identifier != "" || err != nil {
  return o.Identifier, nil
}

// [...]
```

pkg/apis/core/v1/location_genclient.go
```diff
@@ -19,5 +22,28 @@ func (l *Location) EndpointURL(ctx context.Context) (*url.URL, error) {
...
+// GetIdentifier returns the objects identifier
+func (l *Location) GetIdentifier(ctx context.Context) (string, error) {
+	if l.Identifier != "" {
+		return l.Identifier, nil
+	}
+
+	op, err := types.OperationFromContext(ctx)
+	if err != nil {
+		return "", nil
+	}
```
LittleFox94, line 42: same as for the example

LittleFox94 (changes requested):

LittleFox94 (approved):
LGTM!

:tada:

PR #151, opened 2022-07-18
Implement generic client bindings for Kubernetes service

marioreggiori:
### Description

This PR adds generic client bindings for Kubernetes cluster and node pool resources.

### Checklist

### Community Note


pkg/apis/kubernetes/v1/node_pool_types.go
```diff
@@ -0,0 +1,39 @@
...
+// This resource does not support updates
+type NodePool struct {
+	internal.CommonMethods
+	State *internal.State `json:"state,omitempty"`
+
+	Identifier string  `json:"identifier,omitempty" anxcloud:"identifier"`
+	Name       string  `json:"name,omitempty"`
+	Cluster    Cluster `json:"cluster,omitempty"`
+
+	// TODO: should NodePool support 0 replicas?
```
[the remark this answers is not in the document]
toothstone: IMHO it should, but not supplying this will default to 3

pkg/apis/kubernetes/v1/example_test.go
```diff
@@ -0,0 +1,37 @@
...
+
+	if err := a.Create(context.TODO(), &cluster); err != nil {
+		log.Fatalf("failed to create cluster: %s", err)
+	}
+
+	if err := cluster.AwaitCompletion(context.TODO(), a); err != nil {
+		log.Fatalf("failed to await cluster creation: %s", err)
+	}
+
+	nodePool := NodePool{Name: "example-np-00", Cluster: cluster, Replicas: 1, Memory: 2147483648, DiskSize: 10737418240}
```
toothstone, line 28: I suggest using powers of 2 for those memory and disk size numbers (Gi = 2**30).

CHANGELOG.md
```diff
@@ -26,6 +26,7 @@ Some examples, more below in the actual changelog (newer entries are more likely
 ### Added
 * vsphere/v1: Template bindings and FindNamedTemplate helper added to retrieve templates by name (#148, @marioreggiori)
 * generic client: mock implementation (#139, @marioreggiori)
+* kubernetes/v1: implement generic client bindings for cluster and node_pool resources (#151, @marioreggiori)
```
LittleFox94, line 29: I'd favor "kubernetes/v1: add Cluster and NodePool bindings", but that's a nitpick ^^

pkg/apis/internal/common.go
```diff
@@ -0,0 +1,39 @@
+package internal
```
LittleFox94, line 1: not a fan of this package name **or** the type `CommonMethods` - one of those should say it's about `GS` or `GenericServices`

pkg/apis/internal/common.go
```diff
@@ -0,0 +1,39 @@
...
+		}
+
+		res.Body = io.NopCloser(bytes.NewReader([]byte("{}")))
+		return res, nil
+	}
+	return res, nil
+}
+
+// State describes the current State of a GS object
+type State struct {
```
LittleFox94, line 35: This is the same type as `lbaasv1.State`, right? That one also already implements the fun of "sending to the server has another format than retrieving".

Sadly the actual state values vary by API, so we cannot fully extract that stuff .. I'd love to see this implementing `lbaasv1.StateRetriever` (and moving that interface out of lbaasv1) in some way, maybe API-specific `State` types that embed this generic one but add the methods or a way to retrieve the known state values?

pkg/apis/internal/common.go
LittleFox94: I'd go with a package `pkg/apis/internal/gs`, moving `CommonMethods`, `State` and `StateRetriever` there

pkg/apis/kubernetes/v1/cluster_genclient.go
```diff
@@ -0,0 +1,26 @@
...
+import (
+	"context"
+	"net/url"
+
+	"go.anx.io/go-anxcloud/pkg/api"
+	"go.anx.io/go-anxcloud/pkg/api/types"
+)
+
+// EndpointURL returns the common URL for operations on Cluster resource
+func (c *Cluster) EndpointURL(ctx context.Context) (*url.URL, error) {
```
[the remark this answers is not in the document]
LittleFox94: maybe actually unify this with the `NodePool` one, the only difference is the path in the URL.

pkg/apis/internal/common.go
LittleFox94: maybe call `CommonMethods` something like `GenericService` - this reads better for people knowing what generic services are:

```go
type Cluster struct {
  gs.GenericService

  // [...]
}
```

pkg/apis/kubernetes/v1/cluster_types.go
```diff
@@ -0,0 +1,33 @@
...
+	"go.anx.io/go-anxcloud/pkg/apis/internal"
+)
+
+// anxcloud:object
+
+// Cluster represents a Kubernetes cluster
+// This resource does not support updates
+type Cluster struct {
+	internal.CommonMethods
+	State *internal.State `json:"state,omitempty"`
```
LittleFox94, line 16: do generic services specify the state always being stored in the attribute `State`? We could then add it to `internal.CommonMethods`/`gs.GenericService` instead - at least when we don't need a API-specific `State` type for adding the state values ..

pkg/apis/kubernetes/v1/common.go
```diff
@@ -0,0 +1,56 @@
...
+
+	// ErrStateUnknown is returned if a resource has an unknown state
+	ErrStateUnknown = errors.New("resource has an unknown state")
+)
+
+type objectWithState interface {
+	types.Object
+	getState() internal.State
+}
+
```
LittleFox94, line 28: we already have this with `lbaasv1.StateRetriever` and I'd like to reuse that interface (and move it out of `lbaasv1`)

pkg/apis/kubernetes/v1/common.go
```diff
@@ -0,0 +1,56 @@
...
+		}
+
+		select {
+		case <-ctx.Done():
+			return ctx.Err()
+		case <-ticker.C:
+			continue
+		}
+	}
+}
```
LittleFox94, line 56: I like this, but maybe we could allow users to call this directly with any Object that implements `(lbaasv1).StateRetriever`

pkg/apis/kubernetes/v1/cluster_types.go
```diff
@@ -0,0 +1,33 @@
...
+}
+
+func (c *Cluster) getState() internal.State {
+	return *c.State
+}
+
+// AwaitCompletion blocks until the Cluster state is "OK"
+func (c *Cluster) AwaitCompletion(ctx context.Context, a api.API) error {
+	return awaitCompletion(ctx, a, c)
+}
```
LittleFox94, line 33: We don't need this as a method when allowing `awaitCompletion` to be called for any Object that implements `StateRetriever`

pkg/apis/kubernetes/v1/example_test.go
```diff
@@ -0,0 +1,37 @@
...
+import (
+	"context"
+	"log"
+
+	"go.anx.io/go-anxcloud/pkg/api"
+	"go.anx.io/go-anxcloud/pkg/client"
+	"go.anx.io/go-anxcloud/pkg/utils/pointer"
+)
+
+func Example() {
```
LittleFox94, line 12: will `go test` run this example? This would fail when not having `ANEXIA_TOKEN` set, right?

pkg/apis/kubernetes/v1/node_pool_types.go
```diff
@@ -0,0 +1,57 @@
...
+// This resource does not support updates
+type NodePool struct {
+	internal.CommonMethods
+	State *internal.State `json:"state,omitempty"`
+
+	Identifier string  `json:"identifier,omitempty" anxcloud:"identifier"`
+	Name       string  `json:"name,omitempty"`
+	Cluster    Cluster `json:"cluster,omitempty"`
+
+	// Number of replicas can be changed via machine controller. Default: 3 (see FAQ for more details)
```
LittleFox94, line 24: @toothstone  we don't yet have a link to the FAQ we could add here, right?

pkg/apis/kubernetes/v1/node_pool_types.go
```diff
@@ -0,0 +1,57 @@
...
+	if err := json.Unmarshal(b, &val); err != nil {
+		return fmt.Errorf("failed to unmarshal value: %w", err)
+	}
+	*v = gibibyte(val / gibiFactor)
+	return nil
+}
+
+func (v gibibyte) MarshalJSON() ([]byte, error) {
+	return json.Marshal(int(v) * gibiFactor)
+}
```
LittleFox94, line 48: do I read this correctly as "v is given in GiB, so we multiply with gibifactor to send Bytes to the API"? Why not using bytes in the first place then?

pkg/apis/kubernetes/v1/node_pool_types.go
```diff
@@ -0,0 +1,57 @@
...
+
+func (np *NodePool) getState() internal.State {
+	return *np.State
+}
+
+// AwaitCompletion blocks until the NodePool state is "OK"
+func (np *NodePool) AwaitCompletion(ctx context.Context, a api.API) error {
+	return awaitCompletion(ctx, a, np)
+}
```
LittleFox94, line 57: same as for `Cluster`

LittleFox94 (changes requested):
If we are going with extracting `lbaasv1.StateRetriever` into a common `gs` package, we should do this in another pull request - but I think that makes the interface a lot cleaner and more consistent.

pkg/apis/kubernetes/v1/cluster_types.go
toothstone: Generic Service resource definitions have to include one attribute of type State, but it does not have to be called `state`.

pkg/apis/kubernetes/v1/common.go
marioreggiori: I don't think that this is possible, as the function/method would need access to the Object which is not accessible from the embedded type, right?

pkg/apis/kubernetes/v1/example_test.go
marioreggiori: This example does not run with `go test`, as it does not have the "output" comment. Yes, AuthFromEnv returns an error if the token is not set.

pkg/apis/internal/gs/common.go
```diff
@@ -0,0 +1,59 @@
...
+	"net/http"
+
+	"go.anx.io/go-anxcloud/pkg/api/types"
+)
+
+// GenericService has methods overridden for all GS objects in the same way
+type GenericService struct{}
+
+// FilterAPIResponse replaces the API response with an empty json object
+func (cm *GenericService) FilterAPIResponse(ctx context.Context, res *http.Response) (*http.Response, error) {
```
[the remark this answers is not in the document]
marioreggiori: This will be resolved by moving the internal/gs package into a separate PR and using this package with the LBaaS bindings.

pkg/apis/kubernetes/v1/integration_test.go
```diff
@@ -0,0 +1,6 @@
+//go:build integration
+// +build integration
+
+package v1
+
+// TODO: implement integration tests
```
LittleFox94, line 6: the fun thing with your mock tests: change the client to go to the real API and you have an integration test :)

pkg/apis/kubernetes/v1/node_pool_types.go
```diff
@@ -0,0 +1,34 @@
...
+
+	"go.anx.io/go-anxcloud/pkg/api"
+	"go.anx.io/go-anxcloud/pkg/apis/internal/gs"
+)
+
+// anxcloud:object:hooks=RequestBodyHook
+
+// NodePool represents a Kubernetes node pool
+// This resource does not support updates
+type NodePool struct {
```
LittleFox94, line 14: document the attributes please

pkg/apis/internal/gs/common.go
```diff
@@ -0,0 +1,59 @@
...
+		return res, nil
+	}
+	return res, nil
+}
+
+// CommonRequestBody is embedded in the request body types of GS objects
+type CommonRequestBody struct {
+	// this allows us to provide the state as string
+	State string `json:"state,omitempty"`
+}
```
LittleFox94, line 38: this won't work: https://github.com/anexia-it/go-anxcloud/pull/151#discussion_r923376354

LittleFox94 (changes requested):

LittleFox94:
As a note for people stumbling over this:
Our Kubernetes service is still in beta and there is a slight API change coming up soon, so we have this on-hold until then.

pkg/apis/kubernetes/v1/cluster_types.go
```diff
@@ -0,0 +1,37 @@
...
+	"go.anx.io/go-anxcloud/pkg/api"
+	corev1 "go.anx.io/go-anxcloud/pkg/apis/core/v1"
+	"go.anx.io/go-anxcloud/pkg/apis/internal/gs"
+)
+
+// anxcloud:object
+
+// Cluster represents a Kubernetes cluster
+// This resource does not support updates
+type Cluster struct {
```
LittleFox94, line 15: We should add the new service VM fields for LBaaS and NAT gateway

pkg/apis/kubernetes/v1/common.go
```diff
@@ -0,0 +1,80 @@
...
+	"context"
+	"net/url"
+
+	"go.anx.io/go-anxcloud/pkg/api"
+	"go.anx.io/go-anxcloud/pkg/api/types"
+	"go.anx.io/go-anxcloud/pkg/apis/internal/gs"
+	"go.anx.io/go-anxcloud/pkg/utils/object/filter"
+)
+
+func endpointURL(ctx context.Context, o types.Object, apiPath string) (*url.URL, error) {
```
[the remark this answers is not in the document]
LittleFox94: you could combine `OperationFromContext` and `op == OperationUpdate` into a "context check" function to get around this warning

not sure if worth it, tho

pkg/apis/kubernetes/v1/common.go
```diff
@@ -0,0 +1,78 @@
...
+// HasState can be embedded to add the state object to a resource
+type HasState struct {
+	State gs.State `json:"state"`
+}
+
+// StateSuccess checks if the state is one of the successful ones
+func (hs HasState) StateSuccess() bool { return hs.State.ID == "0" }
+
+// StateProgressing checks if the state is marking any change currently being applied
+func (hs HasState) StateProgressing() bool { return hs.State.ID == "2" }
```
LittleFox94, line 53: We should add `3` as a Progressing state

pkg/apis/kubernetes/v1/crud_test.go
```diff
@@ -0,0 +1,398 @@
...
+					Expect(srv.ReceivedRequests()).To(HaveLen(3))
+				}
+			})
+
+			It("returns an error when updating the state fails", func() {
+				// Updating a Cluster is currently not supported at all
+				cluster := Cluster{}
+				err := cluster.AwaitCompletion(context.TODO(), a)
+				Expect(err).To(MatchError(types.ErrUnidentifiedObject))
+			})
```
LittleFox94, line 100: This could actually error-out before you return that error in `EndpointURL` - please always check for your expected error, `api.ErrOperationNotSupported` in this case

pkg/apis/kubernetes/v1/crud_test.go
```diff
@@ -0,0 +1,398 @@
...
+					partialNodePool{"id-1", "name-1"},
+					partialNodePool{"id-2", "name-2"},
+				)
+
+				for i := 0; i < 3; i++ {
+					appendGetNodePoolHandler(srv, fmt.Sprintf("id-%d", i), http.StatusOK, mockedNodePoolResponse(NodePool{Identifier: fmt.Sprintf("id-%d", i), Name: fmt.Sprintf("name-%d", i)}))
+				}
+
+				var pi types.PageInfo
+				err := a.List(context.TODO(), &NodePool{}, api.Paged(1, 1, &pi), api.FullObjects(true))
```
LittleFox94, line 287: you set `limit=1`, have a single `pi.Next()` call and get all three results? That shouldn't work and is probably only working because the mock pagination does not care about the limit?

pkg/apis/kubernetes/v1/crud_test.go
```diff
@@ -0,0 +1,398 @@
...
+
+				for i := 0; i < 3; i++ {
+					appendGetClusterHandler(srv, fmt.Sprintf("id-%d", i), http.StatusOK, mockedClusterResponse(Cluster{
+						Identifier: fmt.Sprintf("id-%d", i),
+						Name:       fmt.Sprintf("name-%d", i),
+					}))
+				}
+
+				var pi types.PageInfo
+				err := a.List(context.TODO(), &Cluster{}, api.Paged(1, 1, &pi), api.FullObjects(true))
```
LittleFox94, line 178: same setting `limit=1` as below for `NodePool`s

pkg/apis/kubernetes/v1/crud_test.go
```diff
@@ -0,0 +1,398 @@
...
+				Expect(nodePools[1].Name).To(Equal("name-1"))
+				Expect(nodePools[2].Name).To(Equal("name-2"))
+			})
+
+			It("can filter by cluster", func() {
+				np := &NodePool{Cluster: Cluster{Identifier: "foo"}}
+				url, err := np.EndpointURL(types.ContextWithOperation(context.TODO(), types.OperationList))
+				Expect(err).ToNot(HaveOccurred())
+				Expect(url.Query().Encode()).To(Equal("filters=cluster%3Dfoo"))
+			})
```
LittleFox94, line 306: this should be moved outside `crud_test` - perfect small unit test without any requirements

LittleFox94 (changes requested):
Please split up the tests a bit more, things that don't actually need anything could go in per-Object `_test.go` files, leaving the big (optionally mocked) integration test in `crud_test.go`

pkg/apis/internal/gs/completion.go
```diff
@@ -0,0 +1,47 @@
...
+
+	"go.anx.io/go-anxcloud/pkg/api"
+)
+
+var (
+	// ErrStateError is returned if a resource could not be provisioned (state "Error")
+	ErrStateError = errors.New("resource has state \"Error\"")
+
+	// ErrStateUnknown is returned if a resource has an unknown state
+	ErrStateUnknown = errors.New("resource has an unknown state")
```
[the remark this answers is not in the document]
marioreggiori: Even though it does not contain double-quotes?

pkg/apis/kubernetes/v1/crud_test.go
marioreggiori: The comment was wrong & misleading. We actually want to check for `types.ErrUnidentifiedObject` in this case.

pkg/apis/internal/gs/completion.go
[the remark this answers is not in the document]
LittleFox94: hm. no. no clue why I wrote that o.O

pkg/apis/kubernetes/v1/example_test.go
LittleFox94: this is still to do @marioreggiori

marioreggiori: @LittleFox94 @toothstone wasn't aware that Golang does not have the ** (pow) expression and decided against an additional import (math pkg).

pkg/apis/kubernetes/v1/unit_test.go
```diff
@@ -0,0 +1,28 @@
...
+
+	"go.anx.io/go-anxcloud/pkg/api/types"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+)
+
+const isIntegrationTest = false
+
+var _ = Describe("Unit Tests", Ordered, func() {
```
LittleFox94, line 17: please don't `Describe("Unit Tests")`, in this case it's probably `Describe("NodePool resource") / It("can filter by cluster")`

LittleFox94 (changes requested):
please also check the older unresolved discussions, there is quite some stuff left I fear

pkg/apis/internal/gs/common.go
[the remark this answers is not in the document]
LittleFox94: when will this separate PR be created?

pkg/apis/kubernetes/v1/node_pool_types.go
```diff
@@ -0,0 +1,43 @@
...
+	// Number of computation cores for each node. The provided cores will be "performance" type CPUs. Must be at least 1 and no more than 16
+	CPUs int `json:"cpus,omitempty"`
+	// RAM size for each node in bytes. Must be a multiple of 1 GiB, at least 2 GiB and no more than 64 GiB
+	Memory int `json:"memory,omitempty"`
+	// Size of the disk for each node in bytes. Its performance type will be the default Anexia Engine provides for the given location.
+	// Must be a multiple of 1 GiB, at least 10 GiB and no more than 1600 GiB
+	DiskSize int `json:"disk_size,omitempty"`
+
+	// Operating system for deployment on the nodes. Default: Flatcar Linux
+	OperatingSystem string `json:"operating_system,omitempty"`
```
LittleFox94, line 37: should this maybe be a "string-enum" type with the supported operating systems?

```go
type OperatingSystem string

const (
    FlatcarLinux OperatingSystem = "Flatcar Linux"
)
```

pkg/apis/kubernetes/v1/node_pool_types.go
```diff
@@ -0,0 +1,43 @@
...
+}
+
+// AwaitCompletion blocks until the NodePool state is "OK"
+func (np *NodePool) AwaitCompletion(ctx context.Context, a api.API) error {
+	return gs.AwaitCompletion(ctx, a, np)
+}
```
LittleFox94, line 43: why this method here? First, users can just call `gs.AwaitCompletion(ctx, a, &np)` and we afaik don't have the method on LBaaS types and second, `gs.HasState` could have this method and all objects would automatically have it?

Originally we thought that's not possible since different state IDs and stuff, but we are going by state type now anyway.

pkg/apis/kubernetes/v1/cluster_types.go
```diff
@@ -0,0 +1,44 @@
...
+	EnableLBaaS *bool `json:"enable_lbaas,omitempty"`
+
+	// Contains a kubeconfig if available
+	KubeConfig string `json:"kubeconfig,omitempty"`
+}
+
+// AwaitCompletion blocks until the Cluster state is "OK"
+func (c *Cluster) AwaitCompletion(ctx context.Context, a api.API) error {
+	return gs.AwaitCompletion(ctx, a, c)
+}
```
LittleFox94, line 44: same question as on NodePool

pkg/apis/kubernetes/v1/cluster_types.go
```diff
@@ -0,0 +1,44 @@
...
+	NeedsServiceVMs *bool `json:"needs_service_vms,omitempty"`
+	// If enabled, Service VMs are configured as NAT gateways connecting the internal cluster network to the internet.
+	// Requires Service VMs.
+	EnableNATGateways *bool `json:"enable_nat_gateways,omitempty"`
+	// If enabled, Service VMs are set up as LBaaS hosts enabling K8s services of type LoadBalancer.
+	// Requires Service VMs.
+	EnableLBaaS *bool `json:"enable_lbaas,omitempty"`
+
+	// Contains a kubeconfig if available
+	KubeConfig string `json:"kubeconfig,omitempty"`
```
LittleFox94, line 38: should be a `*string`, shouldn't it?

pkg/apis/kubernetes/v1/crud_test.go
```diff
@@ -0,0 +1,390 @@
...
+		nodePoolIdentifier string
+
+		clusterName  = "go-anxcloud-" + testutils.RandomHostname()
+		nodePoolName = "go-anxcloud-" + testutils.RandomHostname()
+		location     = corev1.Location{Identifier: "52b5f6b2fd3a4a7eaaedf1a7c019e9ea"} // ANX04
+	)
+
+	mockStateOK := map[string]interface{}{"id": "0"}
+	mockStateError := map[string]interface{}{"id": "1"}
+	mockStatePending := map[string]interface{}{"id": "2"}
```
LittleFox94, line 41: you probably need to mock `"type"` as well

pkg/apis/kubernetes/v1/crud_test.go
```diff
@@ -0,0 +1,390 @@
...
+				cluster := Cluster{}
+				err := cluster.AwaitCompletion(context.TODO(), a)
+				Expect(err).To(MatchError(types.ErrUnidentifiedObject))
+			})
+
+			Context("mock test", Ordered, func() {
+				BeforeEach(func() {
+					if isIntegrationTest {
+						Skip("Can't be tested with integration tests")
+					}
```
LittleFox94, line 105: sure it can, just not make it use the real client?

pkg/apis/kubernetes/v1/crud_test.go
```diff
@@ -0,0 +1,390 @@
...
+				err := a.Create(context.TODO(), &cluster)
+				Expect(err).ToNot(HaveOccurred())
+				Expect(cluster.Identifier).NotTo(BeEmpty())
+				clusterIdentifier = cluster.Identifier
+				Expect(cluster.StatePending()).To(BeTrue())
+				Expect(pointer.BoolVal(cluster.NeedsServiceVMs)).To(BeTrue()) // Default if not set on Create
+			})
+		})
+
+		Context("AwaitCompletion", Ordered, func() {
```
LittleFox94, line 79: this full block could move into unit tests, not needing the Engine to test at all - all Engine stuff is tested with other specs here (like testing `Get`).

LittleFox94 (changes requested):

pkg/apis/kubernetes/v1/node_pool_types.go
marioreggiori: `gs.HasState` would work now as it is no longer internal :+1:
Adding the method to HasState does not work, as we don't have the objects identifier accessible.

pkg/apis/kubernetes/v1/crud_test.go
marioreggiori: I'm a little confused :D
We have a separate run where `isIntegrationTest` is false and in this case we are testing with the mock client. Can you please elaborate?

LittleFox94: Our tests are two modes: unit tests only and all tests (including integration tests) - and I think this clears up what I mean?

pkg/apis/kubernetes/v1/unit_test.go
```diff
@@ -0,0 +1,104 @@
...
+		appendGetClusterHandler(srv, clusterIdentifier, http.StatusOK, map[string]interface{}{"state": mockStateOK})
+
+		cluster := Cluster{Identifier: clusterIdentifier}
+		err := gs.AwaitCompletion(context.TODO(), a, &cluster)
+		Expect(err).ToNot(HaveOccurred())
+		Expect(cluster.StateOK()).To(BeTrue())
+
+		if !isIntegrationTest {
+			Expect(srv.ReceivedRequests()).To(HaveLen(3))
+		}
```
LittleFox94, line 66: is this necessary? It always goes to the mock server?

pkg/apis/kubernetes/v1/unit_test.go
```diff
@@ -0,0 +1,104 @@
...
+	})
+
+	It("returns an error when updating the state fails", func() {
+		cluster := Cluster{}
+		err := gs.AwaitCompletion(context.TODO(), a, &cluster)
+		Expect(err).To(MatchError(types.ErrUnidentifiedObject))
+	})
+
+	Context("mock test", Ordered, func() {
```
LittleFox94, line 75: what is this context for, all tests in `Describe("AwaitCompletion")` are mocked already?

pkg/apis/kubernetes/v1/crud_test.go
```diff
@@ -0,0 +1,334 @@
...
+		clusterIdentifier  string
+		nodePoolIdentifier string
+
+		clusterName  = "go-anxcloud-" + testutils.RandomHostname()
+		nodePoolName = "go-anxcloud-" + testutils.RandomHostname()
+		location     = corev1.Location{Identifier: "52b5f6b2fd3a4a7eaaedf1a7c019e9ea"} // ANX04
+	)
+
+	BeforeEach(func() {
+		srv = ghttp.NewServer()
```
LittleFox94, line 41: this is only needed for non-integration tests, correct?

marioreggiori: Correct. Moved to be only initialized on mock tests

LittleFox94 (approved):

PR #167, opened 2022-09-07
Implement AutoTag option for api.Create

marioreggiori:
### Checklist

### Community Note

### DIFF
--- a/pkg/api/api_implementation.go
+++ b/pkg/api/api_implementation.go
@@ -16,7 +16,7 @@ import (
 	"github.com/go-logr/logr"
 
 	"go.anx.io/go-anxcloud/pkg/api/types"
-	corev1 "go.anx.io/go-anxcloud/pkg/apis/core/v1"
+	corev1helper "go.anx.io/go-anxcloud/pkg/apis/core/v1/helper"
 	"go.anx.io/go-anxcloud/pkg/client"
 )
 
@@ -97,8 +97,20 @@ func (a defaultAPI) Create(ctx context.Context, o types.Object, opts ...types.Cr
 		return fmt.Errorf("failed to execute request: %w", err)
 	}
 
-	if err := corev1.Tag(ctx, a, o, options.AutoTags...); err != nil {
-		return fmt.Errorf("failed to auto tag resource: %w", err)
+	if err := a.handlePostCreateOptions(ctx, o, options); err != nil {
+		return fmt.Errorf("failed running post create options: %w", err)
+	}
+
+	return nil
+}
+
+// handlePostCreateOptions executes configured Create options
+// which should be handled after the object was successfully created
+func (a defaultAPI) handlePostCreateOptions(ctx context.Context, o types.IdentifiedObject, options types.CreateOptions) error {
+	if options.AutoTags != nil {
+		if err := corev1helper.TaggerImplementation.Tag(ctx, a, o, options.AutoTags...); err != nil {
+			return newErrTaggingFailed(err)
+		}
 	}
 
 	return nil
--- a/pkg/api/api_implementation.go
+++ b/pkg/api/api_implementation.go
@@ -94,14 +94,10 @@ func (a defaultAPI) Create(ctx context.Context, o types.Object, opts ...types.Cr
 	}
 
 	if err := a.do(ctx, o, o, &options, types.OperationCreate); err != nil {
-		return fmt.Errorf("failed to execute request: %w", err)
+		return fmt.Errorf("API request failed: %w", err)
 	}
 
-	if err := a.handlePostCreateOptions(ctx, o, options); err != nil {
-		return fmt.Errorf("failed running post create options: %w", err)
-	}
-
-	return nil
+	return a.handlePostCreateOptions(ctx, o, options)
 }
 
 // handlePostCreateOptions executes configured Create options
--- a/pkg/api/api_implementation.go
+++ b/pkg/api/api_implementation.go
@@ -16,6 +16,7 @@ import (
 	"github.com/go-logr/logr"
 
 	"go.anx.io/go-anxcloud/pkg/api/types"
+	corev1 "go.anx.io/go-anxcloud/pkg/apis/core/v1"
 	"go.anx.io/go-anxcloud/pkg/client"
 )
 
@@ -92,7 +93,15 @@ func (a defaultAPI) Create(ctx context.Context, o types.Object, opts ...types.Cr
 		opt.ApplyToCreate(&options)
 	}
 
-	return a.do(ctx, o, o, &options, types.OperationCreate)
+	if err := a.do(ctx, o, o, &options, types.OperationCreate); err != nil {
+		return fmt.Errorf("failed to execute request: %w", err)
+	}
+
+	if err := corev1.Tag(ctx, a, o, options.AutoTags...); err != nil {
+		return fmt.Errorf("failed to auto tag resource: %w", err)
+	}
+
+	return nil
 }
 
 // Update the object on the engine.
--- a/pkg/api/e2e_test.go
+++ b/pkg/api/e2e_test.go
@@ -1,0 +1,93 @@
+//go:build integration
+// +build integration
+
+package api_test
+
+import (
+	"context"
+	"errors"
+	"time"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	"go.anx.io/go-anxcloud/pkg/api"
+	clouddnsv1 "go.anx.io/go-anxcloud/pkg/apis/clouddns/v1"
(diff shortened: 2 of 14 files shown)


pkg/internal/resource_with_tag.go
```diff
@@ -0,0 +1,72 @@
...
+}
+
+func (rwt *ResourceWithTag) EndpointURL(ctx context.Context) (*url.URL, error) {
+	op, err := types.OperationFromContext(ctx)
+	if err != nil {
+		return nil, err
+	}
+
+	if op != types.OperationCreate && op != types.OperationDestroy {
+		return nil, fmt.Errorf("%s: ResourceWithTag only support Create and Destroy operations", "api.ErrOperationNotSupported")
```
marioreggiori, line 34: Maybe we could move the errors in `pkg/api` to `pkg/api/errors`?

pkg/internal/resource_with_tag.go
```diff
@@ -0,0 +1,72 @@
+package internal
```
marioreggiori, line 1: Not sure about this files location. Unfortunately within `pkg/apis/core/v1/internal` the `AutoTag` exec part wouldn't have access to the `ResourceWithTag` struct.

pkg/apis/core/v1/resource_types.go
```diff
@@ -19,13 +23,5 @@ type Resource struct {
...
-
 // ResourceWithTag is a virtual Object used to add (Create) or remove (Destroy) a tag to/from a Resource.
-type ResourceWithTag struct {
-	// Identifier of the Resource which tags to change
-	Identifier string `anxcloud:"identifier"`
-
-	// Name of the Tag to add or remove from the resource
-	Tag string
-}
+type ResourceWithTag = internal.ResourceWithTag
```
marioreggiori, line 27: Not a big fan of 'aliasing internals'.. But maybe it's okay, since we have `Tag`, `Untag` & `ListTags` helper functions in `corev1`? I couldn't find any other solution to resolve import cycles.

marioreggiori:
@LittleFox94 your suggestion to extract the api implementation into its own package instead of the type, allows a much cleaner AutoTag implementation. :+1:

Tests are currently failing, because I still need to rewrite some things to make `api.HTTPError` usable outside of the `api` package (probably just a helper (constructor) function). Also our `xxgenerated_object_test.go` generator needs some work to allow `v1_test` package because corev1 cannot import the `api/implementation` package.

marioreggiori:
Addendum to my previous comment:
We could also move non-generated test into `pkg/apis/core/v1/tests` to avoid changes to the code generator.

I just realized that moving the implementation into a separate package is a breaking change as it cannot be imported by `pkg/api`. Is that something you had in mind @LittleFox94?

LittleFox94:
> Also our xxgenerated_object_test.go generator needs some work to allow v1_test package > because corev1 cannot import the api/implementation package due to import cycles.
> Addendum to my previous comment: We could also move non-generated test into `pkg/apis/core/v1/tests` to avoid changes to the code generator.

We should probably rename all the test packages to `v1_test` and only test exported things. For cases where we have to test internal things, we can still have extra tests in `v1`, but default case is `v1_test` and that's also the only case we generate tests for - does that make sense?

> I just realized that moving the implementation into a separate package is a breaking change as it cannot be imported by `pkg/api`. Is that something you had in mind @LittleFox94?

So `api/implementation` is importing `api` to
* have the type returned by NewAPI `api.API`
  - I'd like that function in `api` instead, with the option-resolving code extracted to make clear what is implementation-specific and what isn't (option-resolving probably isn't)
* have some errors defined in `api`
  - we could move them to `api/types` and re-export them in `api`
  - would also make it easier to use them in api packages - most probably only need to import `api/types` then?
* have the `Option`s in tests
  - all the request options are defined in `api/types`, only the functions to use them are in `api` - maybe we could move them to `api/types` and, again, alias them
* .. I stopped searching at this point

This is a big clusterfuck, especially for a not-that-big feature to add .. let's push that to background processing in our minds for today and look at this again tomorrow?

pkg/apis/core/v1/resource_tags.go
```diff
@@ -72,3 +73,21 @@ func ListTags(ctx context.Context, a api.API, obj types.IdentifiedObject) ([]str
...
+	return Untag(ctx, a, obj, tags...)
+}
+
+func (ti taggerImplementation) ListTags(ctx context.Context, a types.API, obj types.IdentifiedObject) ([]string, error) {
+	return ListTags(ctx, a, obj)
+}
+
+func init() {
+	helper.TaggerImplementation = taggerImplementation(42)
+}
```
LittleFox94, line 93: maybe we should add some comments here, at least this being a workaround and why it is needed

pkg/api/api_implementation.go
```diff
@@ -92,7 +93,27 @@ func (a defaultAPI) Create(ctx context.Context, o types.Object, opts ...types.Cr
 		opt.ApplyToCreate(&options)
 	}
 
-	return a.do(ctx, o, o, &options, types.OperationCreate)
+	if err := a.do(ctx, o, o, &options, types.OperationCreate); err != nil {
+		return fmt.Errorf("failed to execute request: %w", err)
```
LittleFox94, line 97: "API request failed"

pkg/api/api_implementation.go
```diff
@@ -92,7 +93,27 @@ func (a defaultAPI) Create(ctx context.Context, o types.Object, opts ...types.Cr
...
+	}
+
+	if err := a.handlePostCreateOptions(ctx, o, options); err != nil {
+		return fmt.Errorf("failed running post create options: %w", err)
```
LittleFox94, line 101: let's not wrap the error here with another message, instead return a good message from `a.handlePostCreateOptions()` (you already return a good message)

pkg/api/e2e_test.go
```diff
@@ -0,0 +1,93 @@
...
+	BeforeEach(func() {
+		var err error
+		a, err = api.NewAPI(
+			api.WithClientOptions(client.AuthFromEnv(false)),
+		)
+		Expect(err).ToNot(HaveOccurred())
+	})
+
+	Context("AutoTag", func() {
+		It("can automatically tag resources on api.Create", func() {
```
LittleFox94, line 39: maybe we should make an example test from this test case?

LittleFox94 (changes requested):
some small changes and an idea for improvement :)

pkg/api/e2e_test.go
```diff
@@ -0,0 +1,93 @@
...
+
+			DeferCleanup(func() {
+				_ = a.Destroy(context.TODO(), &zone)
+			})
+
+			err := a.Create(context.TODO(), &zone, api.AutoTag("foo", "bar", "baz"))
+			Expect(err).To(HaveOccurred())
+			var e *api.ErrTaggingFailed
+			Expect(errors.As(err, &e)).To(BeTrue())
+		})
```
LittleFox94, line 91: this fails because we cannot tag CloudDNS zones, correct? We should add a comment here why we expect this to fail - just to make that clear for future readers :)

pkg/api/e2e_test.go
```diff
@@ -0,0 +1,93 @@
...
+				DescriptionCustomer: "go-anxcloud test api.Create AutoTag " + testutils.RandomHostname(),
+				Locations: []corev1.Location{
+					{Identifier: "52b5f6b2fd3a4a7eaaedf1a7c019e9ea"},
+				},
+			}
+
+			DeferCleanup(func() {
+				if vlan.Identifier == "" {
+					return
+				}
```
LittleFox94, line 50: ~we can go without this, as we only clean up in `DeferCleanup`~

ah wait, we need this for when VLAN creation fails

pkg/api/e2e_test.go
LittleFox94: thanks - another question: with our recently added custom identifiers feature, could we add support for the resource identifier to CloudDNS now?

LittleFox94 (approved):
left a question, but that doesn't stand in the way of merging this :)

pkg/api/e2e_test.go
marioreggiori: Not without using corev1.Resource because CloudDNS does not expose the common identifier at all.

LittleFox94: yeah, was thinking of adding an `Identifier` attribute to the Zone type, which is empty most of the time and have our custom identifier handling return the name for all operations - except for tagging it would take that identifier.

With that, we could at least manually set the identifier and tag a Zone and are prepared for Engine CDNS API returning the Identifier.

marioreggiori: We'd probably need to add an additional operation to make this work. Manual tagging of zones can be done by using `corev1.Tag(ctx, apiClient, &corev1.Resource{Identifier: "<common zone identifier>"})`.

PR #208, opened 2023-01-27
Extend Kubernetes Cluster bindings to include prefixes

marioreggiori:
### Description

This PR adds fields to the Kubernetes cluster API resource which allow configurable prefixes. It also adds common resource types to work with generic services.

### Checklist

### Community Note

### DIFF
--- a/pkg/apis/common/gs/resource_test.go
+++ b/pkg/apis/common/gs/resource_test.go
@@ -1,0 +1,26 @@
+package gs
+
+import (
+	"encoding/json"
+	"testing"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+)
+
+func TestCommonGenericServiceAPIResources(t *testing.T) {
+	RegisterFailHandler(Fail)
+	RunSpecs(t, "test suite for common gs API resources")
+}
+
+var _ = DescribeTable("PartialResourceList",
+	func(list PartialResourceList, expected string) {
+		data, err := json.Marshal(&list)
+
+		Expect(err).ToNot(HaveOccurred())
+		Expect(string(data)).To(Equal(expected))
+	},
+	Entry("empty list", PartialResourceList{}, `""`),
+	Entry("single resource", PartialResourceList{{Identifier: "foo"}}, `"foo"`),
+	Entry("multiple resources", PartialResourceList{{Identifier: "foo"}, {Identifier: "bar"}}, `"foo,bar"`),
+)
--- a/pkg/apis/common/resource.go
+++ b/pkg/apis/common/resource.go
@@ -1,0 +1,16 @@
+package common
+
+import (
+	"encoding/json"
+)
+
+// PartialResource represents a linked resource
+type PartialResource struct {
+	Identifier string `json:"identifier,omitempty"`
+	Name       string `json:"name,omitempty"`
+}
+
+// MarshalJSON unfolds the resource to its identifier
+func (pr PartialResource) MarshalJSON() ([]byte, error) {
+	return json.Marshal(pr.Identifier)
+}
--- a/pkg/apis/common/resource_test.go
+++ b/pkg/apis/common/resource_test.go
@@ -1,0 +1,24 @@
+package common
+
+import (
+	"encoding/json"
+	"testing"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+)
+
+func TestCommonAPIResources(t *testing.T) {
+	RegisterFailHandler(Fail)
+	RunSpecs(t, "test suite for common API resources")
+}
+
+var _ = Describe("PartialResource", func() {
+	It("json marshalls to the resources identifier", func() {
+		pr := PartialResource{Identifier: "foo", Name: "bar"}
+		data, err := json.Marshal(&pr)
+
+		Expect(err).ToNot(HaveOccurred())
+		Expect(string(data)).To(Equal(`"foo"`))
+	})
+})
--- a/pkg/apis/kubernetes/v1/cluster_genclient.go
+++ b/pkg/apis/kubernetes/v1/cluster_genclient.go
@@ -2,16 +2,51 @@ package v1
 
 import (
 	"context"
+	"errors"
 	"net/url"
+
+	"go.anx.io/go-anxcloud/pkg/api/types"
+	"go.anx.io/go-anxcloud/pkg/utils/pointer"
 )
+
+// ErrManagedPrefixSet is returned if a prefix is both set to managed and also set
+// to an existing prefix on create
+var ErrManagedPrefixSet = errors.New("managed prefixes cannot be set on create")
 
 // EndpointURL returns the common URL for operations on Cluster resource
 func (c *Cluster) EndpointURL(ctx context.Context) (*url.URL, error) {
 	return endpointURL(ctx, c, "/api/kubernetes/v1/cluster.json")
 }
 
+// explicitlyFalse returns true if the value of the provided
+// bool pointer is set to false, nil and true pointer return false
+func explicitlyFalse(b *bool) bool {
+	return b != nil && !pointer.BoolVal(b)
+}
+
+// prefixConfigurationValidCreate validates that prefixes configured as managed
+// are not set. This is only relevant for `Create` operations.
+func prefixConfigurationValidCreate(c *Cluster) bool {
+	var (
+		internalV4 = c.InternalIPv4Prefix == nil || explicitlyFalse(c.ManageInternalIPv4Prefix)
+		externalV4 = c.ExternalIPv4Prefix == nil || explicitlyFalse(c.ManageExternalIPv4Prefix)
+		externalV6 = c.ExternalIPv6Prefix == nil || explicitlyFalse(c.ManageExternalIPv6Prefix)
+	)
+
+	return internalV4 && externalV4 && externalV6
+}
+
 // FilterAPIRequestBody adds the CommonRequestBody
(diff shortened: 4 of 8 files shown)


pkg/apis/kubernetes/v1/cluster_types.go
```diff
@@ -31,6 +32,23 @@ type Cluster struct {
...
 
+	// Identifier of an existing internal v4 prefix in which the cluster should be created.
+	// A new prefix is created if `ManageInternalIPv4Prefix` is true.
+	InternalIPv4Prefix *common.PartialResource `json:"internal_ipv4_prefix,omitempty"`
+	// Identifier of an existing external v4 prefix in which the cluster should be created.
+	// A new prefix is created if `ManageExternalIPv4Prefix` is true.
+	ExternalIPv4Prefix *common.PartialResource `json:"external_ipv4_prefix,omitempty"`
+	// Identifier of an existing external v6 prefix in which the cluster should be created.
+	// A new prefix is created if `ManageExternalIPv6Prefix` is true.
+	ExternalIPv6Prefix *common.PartialResource `json:"external_ipv6_prefix,omitempty"`
```
LittleFox94, line 43: Change docs, these attributes are not only useful when creating a cluster but you will also get the automatically created prefixes after creating a cluster with managed prefixes

Also reword the second part to ensure users set the attributes below to false if they want to supply their own prefixes on cluster creation - "Creating clusters in existing prefixes requires the Manage..Prefix attribute below to be set to false"

pkg/apis/kubernetes/v1/cluster_types.go
```diff
@@ -31,6 +32,23 @@ type Cluster struct {
...
+
+	// If set to true an internal v4 prefix is automatically created for the cluster. Defaults to true if not set.
+	ManageInternalIPv4Prefix *bool `json:"manage_internal_ipv4_prefix,omitempty"`
+	// If set to true an external v4 prefix is automatically created for the cluster. Defaults to true if not set.
+	ManageExternalIPv4Prefix *bool `json:"manage_external_ipv4_prefix,omitempty"`
+	// If set to true an external v6 prefix is automatically created for the cluster. Defaults to true if not set.
+	ManageExternalIPv6Prefix *bool `json:"manage_external_ipv6_prefix,omitempty"`
```
LittleFox94, line 50: maybe we actually want to default them for Create calls in go-anxcloud, based on whether prefixes were configured?

pkg/apis/common/resource.go
```diff
@@ -0,0 +1,30 @@
...
+import (
+	"encoding/json"
+	"strings"
+)
+
+// PartialResource represents a linked resource in GS
+type PartialResource struct {
+	Identifier string `json:"identifier,omitempty"`
+	Name       string `json:"name,omitempty"`
+}
```
LittleFox94, line 12: this not only exists in GS but is a pattern in our APIs quite often - just strip the "in GS" from the docs

pkg/apis/common/resource.go
```diff
@@ -0,0 +1,30 @@
...
+
+// MarshalJSON unfolds the resource to its identifier
+func (pr PartialResource) MarshalJSON() ([]byte, error) {
+	return json.Marshal(pr.Identifier)
+}
+
+// PartialResourceList represents a linked resource list in GS
+type PartialResourceList []PartialResource
```
LittleFox94, line 20: this thing is needed for resource lists represented in the API as a single string, maybe rename the type to `PartialResourceListString` or something?

IIRC there are other cases where the API works with a `[]PartialResource` as expected and I want to prevent people being confused they should always use this type - hence the slightly more specific name

pkg/apis/kubernetes/v1/cluster_types.go
marioreggiori: I think that could lead to confusion, as this would have too many hidden conditions.
If we don't limit the defaulting to Create operations, the following (pseudo!)code could easily "unmanage" prefixes by accident when updates are supported in the future.

```golang
cluster := kubernetesv1.Cluster{Identifier: "foo"}
api.Get(&cluster)
cluster.Name = "bar"
api.Update(&cluster)
```

pkg/apis/kubernetes/v1/cluster_types.go
marioreggiori: Maybe something like
```golang
	// Identifier of an internal v4 prefix (to be) assigned to the cluster. If ManageInternalIPv4Prefix
	// is set to false, the value of this field is used to create the cluster. Otherwise a new prefix
	// will be created automatically. Updates to actual value after Create, Get & Update.
	InternalIPv4Prefix *common.PartialResource `json:"internal_ipv4_prefix,omitempty"`
```

pkg/apis/kubernetes/v1/cluster_types.go
marioreggiori: I think a different default logic on create would cause more confusion. Maybe we should instead return an error if prefix is set and managed on create?

pkg/apis/common/resource.go
marioreggiori: Not sure if `PartialResourceListString` is a better name for this type as it only encodes to (not decodes from) string. Maybe we should move this type to the `gs` sub-pkg?

pkg/apis/kubernetes/v1/cluster_types.go
LittleFox94: Good, but I don't like "the value of this field" as we can be more specific and I don't like the last sentence, what about those?

`[...] the Prefix given in this field is used when creating the cluster, otherwise a new [...]`

`Updates to actual value after Create, Get & Update.` -> `The API will always return the Prefix for the Cluster, when ManageInternalIPv4Prefix is true, this will be the Prefix that was created automatically.`

pkg/apis/kubernetes/v1/cluster_types.go
LittleFox94: Returning an error sounds good!

pkg/apis/common/resource.go
LittleFox94: I think it's *technically* not related to GS, so moving it in the `gs` package feels wrong to me

Any other ideas how to call it, given my reason to rename it above?

marioreggiori: AFAIK encoding to comma separated identifiers is unique to GS. The pyengine also has similar logic in a file called `generic_service_resource.py`. Can you point me to an engine endpoint, which also uses this?

I don't think we should make a (more) general purpose type out of this as it has a specialized Marshaler, and therefore still think it'l be best placed within the gs package.

pkg/apis/common/gs/resource_test.go
```diff
@@ -0,0 +1,27 @@
...
+		data, err := json.Marshal(&list)
+
+		Expect(err).ToNot(HaveOccurred())
+		Expect(string(data)).To(Equal(expected))
+	},
+		Entry("empty list", PartialResourceList{}, `""`),
+		Entry("single resource", PartialResourceList{{Identifier: "foo"}}, `"foo"`),
+		Entry("multiple resources", PartialResourceList{{Identifier: "foo"}, {Identifier: "bar"}}, `"foo,bar"`),
+	)
+})
```
LittleFox94, line 27: can `DescribeTable` be a top-level block? If so strip the `Describe`

pkg/apis/common/resource_test.go
```diff
@@ -0,0 +1,26 @@
...
+	Context("PartialResource", func() {
+		It("json marshalls to the resources identifier", func() {
+			pr := PartialResource{Identifier: "foo", Name: "bar"}
+			data, err := json.Marshal(&pr)
+
+			Expect(err).ToNot(HaveOccurred())
+			Expect(string(data)).To(Equal(`"foo"`))
+		})
+	})
+})
```
LittleFox94, line 26: please `Describe("PartialResource", ...)` instead

pkg/apis/kubernetes/v1/cluster_genclient.go
```diff
@@ -2,16 +2,45 @@ package v1
...
+
 // EndpointURL returns the common URL for operations on Cluster resource
 func (c *Cluster) EndpointURL(ctx context.Context) (*url.URL, error) {
 	return endpointURL(ctx, c, "/api/kubernetes/v1/cluster.json")
 }
 
+// prefixConfigurationValidCreate validates that prefixes configured as managed
+// are not set. This is only relevant for `Create` operations.
+func prefixConfigurationValidCreate(c *Cluster) bool {
+	if (c.InternalIPv4Prefix != nil && c.ManageInternalIPv4Prefix != pointer.Bool(false)) ||
```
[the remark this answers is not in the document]
LittleFox94: `if $bool return true else return false` is a but funny ^^

try a variable per prefix and return `internal_v4 && external_v4 && external_v6`
```go
internal_v4 := c.InternalIPv4Prefix != nil && c.ManageInternalIPv4Prefix != pointer.Bool(false)
```

pkg/apis/common/gs/resource_test.go
```diff
@@ -0,0 +1,25 @@
...
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+)
+
+func TestCommonGenericServiceAPIResources(t *testing.T) {
+	RegisterFailHandler(Fail)
+	RunSpecs(t, "test suite for common gs API resources")
+}
+
+var _ = DescribeTable("PartialResourceList", func(list PartialResourceList, expected string) {
```
LittleFox94, line 16: when you move the func to the next line, the block looks a bit less weird, with the Entries following it

LittleFox94 (approved):

LittleFox94 (approved):

PR #258, opened 2023-05-10
SYSENG-867: Cleanup workflows

11Tuvork28:
### Description

Cleans up workflows and merges:
- `integration.yml`
- `unit.yml`

Into the `code.yml` workflow and removes the `ok-to-test` workflow.
Integration tests are now only run if a PR is labeled with `integration_tests`.
The `code.yml` workflow now only runs if code is changed, meaning that documentation is ignored, further more unit tests always run regardless if jobs before that failed.


eiabea (approved):

.github/workflows/code.yml
```diff
@@ -50,3 +57,58 @@ jobs:
...
+          verbose: true
+  integration-tests:
+    if: ${{ github.event.label.name == 'integration_tests' }}
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        go:
+        - version: "1.19"
+          name: target
+    name: "Integration tests with ${{ matrix.go.name }} Go (ok-to-test or trusted)"
```
LittleFox94, line 93: that name should've been changed

.github/workflows/code.yml
```diff
@@ -50,3 +57,58 @@ jobs:
...
+          name: latest
+    name: "Unit tests with ${{ matrix.go.name }} Go"
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions/setup-go@v3
+        with:
+          go-version: ${{ matrix.go.version }}
+      - name: run unit tests
+        run: make test
+      - name: Upload coverage to Codecov
```
LittleFox94, line 78: could we add the integration test run here, after unit tests but before uploading coverage? Could make some coverage problems go away if we can upload the reports for "target go" and "latest go" in one go each

PR #352, opened 2024-03-19
Refactor GS API bindings

ghost:
### Checklist

### Community Note


pkg/apis/kubernetes/v1/cluster_genclient.go
```diff
@@ -2,59 +2,10 @@ package v1
...
-// are not set. This is only relevant for `Create` operations.
-func prefixConfigurationValidCreate(c *Cluster) bool {
-	var (
-		internalV4 = c.InternalIPv4Prefix == nil || explicitlyFalse(c.ManageInternalIPv4Prefix)
-		externalV4 = c.ExternalIPv4Prefix == nil || explicitlyFalse(c.ManageExternalIPv4Prefix)
-		externalV6 = c.ExternalIPv6Prefix == nil || explicitlyFalse(c.ManageExternalIPv6Prefix)
-	)
-
-	return internalV4 && externalV4 && externalV6
-}
```
anx-mschaefer, line 37: TODO: should probably go into a different hook (maybe a new RequestValidationHook)

pkg/apis/kubernetes/v1/cluster_test.go
```diff
@@ -67,35 +65,3 @@ var _ = Describe("AwaitCompletion", Ordered, func() {
...
-})
-
-var _ = DescribeTable("explicitlyFalse",
-	func(b *bool, expected bool) {
-		Expect(explicitlyFalse(b)).To(Equal(expected))
-	},
-	Entry("explicitly false", pointer.Bool(false), true),
-	Entry("explicitly true", pointer.Bool(true), false),
-	Entry("implicitly false (nil)", nil, false),
-)
```
anx-mschaefer, line 101: TODO: restore with new hook

LittleFox94:
I don't like how we have a ticket to plan this and then there's lots of code already that I have to parse again  ^^`

Short look only rn, as I actually tried to get another hour of sleep but had another idea to solve this, but looks like you add code to the generic client to handle specific cases? I think we talked about this before.. it's not gonna be maintainable with all the different behaviors we have in different APIs, that's why it was hooks in the first place

Our problem is the hooks being attached to the objects and working on the Object they get via the receiver, right?

What if we moved _all_ the hooks away from the object types itself and have them optionally implement an interface "client overrides" where we have those hooks instead?
* they would get the object via an argument, so extending objects via embedding still works
* we still have custom logic per object type if needed but not in the generic client itself
* we can share lots more code since e.g. all of GS has a single "client override" class they can all use
* it makes the object types a bit cleaner  conceptually - they are what's transferred via API request, not the logic for that
* existing objects could just return themselves as their "client overrides" object to still have the hooks directly on themselves (with slightly changed interface), easing migration to this pattern

LittleFox94:
Ah so you are essentially doing those things here:

* add a way to flatten linked objects to their identifier
  - implemented here with reflection in the generic client
  - implementing this with (generated) code in the Objects would be better imho, reflection can get slow and this is something we should be able to do at build-time rather easily?
* remove the need for hooks for GS State
* remove the need for hooks to deal with GS APIs returning the object data on `DELETE`
  - I still think this does not belong into the generic client and should work good with a "client override" class instead of hooks on all the affected Object types

Maybe we should just have `Encode` and `Decode` methods on the "client overrides" interface, getting a `context.Context`, `io.Reader` and `Object` to decode into - they could have different encoding/decoding logic per `operation` and even call out to the Object's `Encode`/`Decode` which is generated to handle e.g. this identifier flattening?

anx-mschaefer:
> implementing this with (generated) code in the Objects would be better imho, reflection can get slow and this is something we should be able to do at build-time rather easily?

I did a benchmark and while flattening has a relative high overhead compared to just json encoding, considering the unit of the extra time needed is not a valid concern for an API client IMHO.
```
BenchmarkEncodeWithFlatten-8              134679              9044 ns/op
BenchmarkEncodeWithoutFlatten-8          5170754               247.3 ns/op
```

Implementing this with generated code would probably not be as trivial as it appears with the needed support for embedded structs.

---

> remove the need for hooks to deal with GS APIs returning the object data on DELETE
> - I still think this does not belong into the generic client and should work good with a "client override" class instead of hooks on all the affected Object types

With the plan to eventually move everything to GS I think it'd be better located in the client core than overriding it for every resource.

PR #361, opened 2024-04-24
SYSENG-1357: generic client default options

anx-mschaefer:
### Description

This PR adds functionality to define default requests options via `WithRequestOptions` when creating a new generic API client.

### Checklist

### Community Note

### DIFF
--- a/pkg/api/api_implementation.go
+++ b/pkg/api/api_implementation.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"io"
 	"mime"
@@ -30,7 +31,8 @@ type defaultAPI struct {
 	client client.Client
 	logger *logr.Logger
 
-	clientOptions []client.Option
+	clientOptions  []client.Option
+	requestOptions []types.Option
 }
 
 // NewAPIOption is the type for giving options to the NewAPI function.
@@ -40,6 +42,13 @@ type NewAPIOption func(*defaultAPI)
 func WithClientOptions(o ...client.Option) NewAPIOption {
 	return func(a *defaultAPI) {
 		a.clientOptions = append(a.clientOptions, o...)
+	}
+}
+
+// WithRequestOptions configures default options applied to requests
+func WithRequestOptions(opts ...types.Option) NewAPIOption {
+	return func(a *defaultAPI) {
+		a.requestOptions = opts
 	}
 }
 
@@ -79,8 +88,12 @@ func NewAPI(opts ...NewAPIOption) (API, error) {
 // Get the identified object from the engine.
 func (a defaultAPI) Get(ctx context.Context, o types.IdentifiedObject, opts ...types.GetOption) error {
 	options := types.GetOptions{}
-	for _, opt := range opts {
-		opt.ApplyToGet(&options)
+	var err error
+	for _, opt := range resolveRequestOptions(a.requestOptions, opts) {
+		err = errors.Join(err, opt.ApplyToGet(&options))
+	}
+	if err != nil {
+		return fmt.Errorf("apply request options: %w", err)
 	}
 
 	return a.do(ctx, o, o, &options, types.OperationGet)
@@ -89,8 +102,12 @@ func (a defaultAPI) Get(ctx context.Context, o types.IdentifiedObject, opts ...t
 // Create the given object on the engine.
 func (a defaultAPI) Create(ctx context.Context, o types.Object, opts ...types.CreateOption) error {
 	options := types.CreateOptions{}
-	for _, opt := range opts {
-		opt.ApplyToCreate(&options)
+	var err error
+	for _, opt := range resolveRequestOptions(a.requestOptions, opts) {
+		err = errors.Join(err, opt.ApplyToCreate(&options))
+	}
+	if err != nil {
+		return fmt.Errorf("apply request options: %w", err)
 	}
 
 	if err := a.do(ctx, o, o, &options, types.OperationCreate); err != nil {
@@ -115,8 +132,12 @@ func (a defaultAPI) handlePostCreateOptions(ctx context.Context, o types.Identif
 // Update the object on the engine.
 func (a defaultAPI) Update(ctx context.Context, o types.IdentifiedObject, opts ...types.UpdateOption) error {
 	options := types.UpdateOptions{}
-	for _, opt := range opts {
-		opt.ApplyToUpdate(&options)
+	var err error
+	for _, opt := range resolveRequestOptions(a.requestOptions, opts) {
+		err = errors.Join(err, opt.ApplyToUpdate(&options))
+	}
+	if err != nil {
+		return fmt.Errorf("apply request options: %w", err)
 	}
 
 	return a.do(ctx, o, o, &options, types.OperationUpdate)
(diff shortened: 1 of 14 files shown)


89Q12 (approved):
LGTM

pkg/api/types/options.go
```diff
@@ -24,33 +24,39 @@ type Options interface {
...
-	ApplyToDestroy(*DestroyOptions)
+	ApplyToDestroy(*DestroyOptions) error
 }
 
+type AnyOption interface {
+	ApplyToAny(Options) error
+}
+
+type Option interface{}
+
```
LittleFox94, line 59: missing doc comments

pkg/apis/kubernetes/v1/environments.go
```diff
@@ -0,0 +1,14 @@
+package v1
+
+import (
+	"go.anx.io/go-anxcloud/pkg/api"
+	"go.anx.io/go-anxcloud/pkg/api/types"
+)
+
+func EnvironmentProduction(override bool) types.Option {
```
LittleFox94, line 8: missing doc comment

pkg/apis/kubernetes/v1/node_pool_test.go
```diff
@@ -12,7 +12,8 @@ import (
 var _ = Describe("NodePool resource", func() {
 	It("can filter by cluster", func() {
 		np := &NodePool{Cluster: Cluster{Identifier: "foo"}}
-		url, err := np.EndpointURL(types.ContextWithOperation(context.TODO(), types.OperationList))
+		url, err := np.EndpointURL(
+			types.ContextWithOperation(types.ContextWithOptions(context.TODO(), &types.ListOptions{}), types.OperationList))
```
LittleFox94, line 16: why is this change necessary?

pkg/api/options.go
```diff
@@ -34,3 +38,32 @@ func FullObjects(fullObjects bool) ListOption {
...
+// ApplyToAny applies the EnvironmentOption to any request type
+func (o EnvironmentOption) ApplyToAny(opts types.Options) error {
+	return opts.Set(fmt.Sprintf("environment/%s", o.APIGroup), o.EnvPathSegment, o.Override)
+}
+
+// GetEnvironmentPathSegment retrieves the environment path segment of a given API group
+// or the provided defaultValue if no environment override is set
+func GetEnvironmentPathSegment(ctx context.Context, apiGroup, defaultValue string) (string, error) {
+	if options, err := types.OptionsFromContext(ctx); err != nil {
+		return "", err
```
LittleFox94, line 59: shouldn't it just return the default value in this case?

pkg/api/options.go
```diff
@@ -34,3 +38,32 @@ func FullObjects(fullObjects bool) ListOption {
...
+	} else if env, err := options.Get(fmt.Sprintf("environment/%s", apiGroup)); errors.Is(err, types.ErrKeyNotSet) {
+		return defaultValue, nil
+	} else if err != nil {
+		return "", err
+	} else if envString, ok := env.(string); !ok {
+		return "", ErrEnvironmentSegmentNotTypeString
```
LittleFox94, line 65: if the comment above is accepted, this is the last branch that could return an error - could it be sensible to have a `environments` variable on `types.commonOptions` that can only hold strings, removing the need for this function to be able to return an error, which then also helps with the now 5 returns in `apis/kubernetes/v1.endpointURL`?

pkg/api/api_implementation.go
```diff
@@ -79,8 +88,19 @@ func NewAPI(opts ...NewAPIOption) (API, error) {
 // Get the identified object from the engine.
 func (a defaultAPI) Get(ctx context.Context, o types.IdentifiedObject, opts ...types.GetOption) error {
 	options := types.GetOptions{}
+	var err error
+	for _, requestOpt := range a.requestOptions {
+		if getOpt, ok := requestOpt.(types.GetOption); ok {
+			err = errors.Join(err, getOpt.ApplyToGet(&options))
```
LittleFox94, line 94: I would've just returned on the first option returning an error, but I guess collecting all of them could be useful in some situations - "ok, would have done differently tho" ^^

LittleFox94 (changes requested):
found a few things that would be worth changing, but please also first note what I wrote you in Mattermost

pkg/api/options.go
anx-mschaefer: Hmm.. not sure about that. I think Options are always expected to be available on the context when using this helper (`GetEnvironmentPathSegment`) defaulting to prod (in most cases) on error sounds a bit dangerous.

pkg/apis/kubernetes/v1/node_pool_test.go
anx-mschaefer: See comment below

pkg/apis/kubernetes/v1/environments.go
anx-mschaefer: Added the comment (not pushed yet).

But do we need this file and its content at all in this project? We are already defaulting to the prod environment.

pkg/api/api_implementation.go
anx-mschaefer: Did that to make codeclimate less angry (too many returns) :grimacing:

pkg/apis/kubernetes/v1/environments.go
LittleFox94: hm probably not, good point - up to you

pkg/api/options.go
LittleFox94: really not sure .. I think the risk to default to prod due to this is pretty low and can be reasonably enough tested with unit tests, but makes the code using this function a lot nicer in exchange

And dealing with the risk is also possible by using tokens that cannot use the wrong service

pkg/api/options.go
LittleFox94: something like that, yea

pkg/api/api_implementation.go
```diff
@@ -79,8 +88,17 @@ func NewAPI(opts ...NewAPIOption) (API, error) {
...
+		}
+	}
 	for _, opt := range opts {
-		opt.ApplyToGet(&options)
+		err = errors.Join(err, opt.ApplyToGet(&options))
+	}
```
LittleFox94, line 99: this can be prettied up a bit for all the request methods with a small generic helper:

```go
func resolveRequestOptions[T any](commonOptions []types.Option, requestOptions []T) []T {
	return append(filterOptions[T](commonOptions), requestOptions...)
}

func filterOptions[T any](opts []types.Option) []T {
	ret := make([]T, 0, len(opts))
	for _, v := range opts {
		if v, ok := v.(T); ok {
			ret = append(ret, v)
		}
	}

	return ret
}
```

With that present, the marked code would be replaced with this:
```go
for _, opt := range resolveRequestOptions(a.requestOptions, opts) {
	err = errors.Join(err, opt.ApplyToGet(&options))
}
```

pkg/api/types/options.go
```diff
@@ -73,3 +107,26 @@ func (o *commonOptions) Set(key string, val interface{}, overwrite bool) error {
 	o.additional[key] = val
 	return nil
 }
+
+func (o commonOptions) GetEnvironment(key string) (string, error) {
```
LittleFox94, line 111: no doc comment..

pkg/api/errors.go
```diff
@@ -22,6 +22,9 @@ var (
 
 	// ErrContextRequired is returned when a nil context was passed as argument.
 	ErrContextRequired = errors.New("no context given")
+
+	// ErrEnvironmentSegmentNotTypeString is returned when the environment segment is not of type string
+	ErrEnvironmentSegmentNotTypeString = errors.New("environment segment is not of type string")
```
LittleFox94, line 27: not needed anymore, right?

89Q12 (commented):
LGTM, the failing integration tests well meh.
Otherwise doc strings have all been added.

LittleFox94:
All relevant integration tests went through, I'm ok with merging without all of them being green

PR #378, opened 2024-06-05
Add ip-version and prefix filter attributes to `ReserveRandom` request

anx-mschaefer:
### Checklist

### Community Note

### DIFF
--- a/pkg/ipam/address/address.go
+++ b/pkg/ipam/address/address.go
@@ -69,11 +69,35 @@ type Create struct {
 	Organization        string `json:"organization"`
 }
 
+// IPReserveVersionLimit limits the IP version for address reservations
+type IPReserveVersionLimit int
+
+const (
+	// IPReserveVersionLimitIPv4 specifies that only IPv4 addresses should be reserved
+	IPReserveVersionLimitIPv4 IPReserveVersionLimit = 4
+
+	// IPReserveVersionLimitIPv4 specifies that only IPv6 addresses should be reserved
+	IPReserveVersionLimitIPv6 IPReserveVersionLimit = 6
+)
+
 // ReserveRandom defines metadata of addresses to reserve randomly.
 type ReserveRandom struct {
 	LocationID string `json:"location_identifier"`
 	VlanID     string `json:"vlan_identifier"`
 	Count      int    `json:"count"`
+
+	// PrefixID limits the potential addresses to a specific prefix
+	// within the specified VLAN. Defaults to all prefixes within the VLAN.
+	PrefixID string `json:"prefix_identifier,omitempty"`
+
+	// IPVersion limits the potential addresses to a specific IP version.
+	// Defaults to v4 or v6, depending on availability.
+	IPVersion IPReserveVersionLimit `json:"ip_version,omitempty"`
+
+	// ReservationPeriod specifies how many seconds the addresses should be reserved.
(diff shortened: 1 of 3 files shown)


pkg/ipam/address/address.go
```diff
@@ -74,6 +74,18 @@ type ReserveRandom struct {
...
+
+	// PrefixID limits the potential addresses to a specific prefix
+	// within the specified VLAN. Defaults to all prefixes within the VLAN.
+	PrefixID string `json:"prefix_identifier,omitempty"`
+
+	// IPVersion limits the potential addresses to a specific IP version.
+	// Defaults to v4 & v6.
+	IPVersion int `json:"ip_version,omitempty"`
+
+	// ReservationPeriod specifies how many seconds the addresses should be reserved.
```
nachtjasmin, line 86: Two questions come to my mind:

1. What happens if we pass a negative value?
2. What happens after the reservation period? If I assign the IP to a VM, is it going to be removed after the period again?

I think a sentence like "If the IP address hasn't been assigned to a resource within the period, it is released again." could help here.

pkg/ipam/address/address.go
```diff
@@ -74,6 +74,18 @@ type ReserveRandom struct {
...
 	VlanID     string `json:"vlan_identifier"`
 	Count      int    `json:"count"`
+
+	// PrefixID limits the potential addresses to a specific prefix
+	// within the specified VLAN. Defaults to all prefixes within the VLAN.
+	PrefixID string `json:"prefix_identifier,omitempty"`
+
+	// IPVersion limits the potential addresses to a specific IP version.
+	// Defaults to v4 & v6.
+	IPVersion int `json:"ip_version,omitempty"`
```
nachtjasmin, line 84: So, `0` means IPv4 and IPv6, `4` means IPv4 and `6` means IPv6. That's one hell of an API (not to blame it on you, it's not your fault.)

But I think we should at least add some validation *and* better documentation to let users know what's going to happen at which point. Maybe we could even add a "enum" type for this property to have at least some self-documented code. Something along the lines of:

```go
type IPVersionLimit int

const (
	IPVersionLimitDualstack IPVersionLimit = 0
	IPVersionLimitIPv4 IPVersionLimit = 4
	IPVersionLimitIPv6 IPVersionLimit = 6 
)
```

What do you think?

nachtjasmin (changes requested):
I think the documentation could be a bit better, but overall I'm fine with it :+1:

pkg/ipam/address/address.go
anx-mschaefer: 1. Changed to `uint`
2. Will add the proposed comment (changed to plural "addresses")

pkg/ipam/address/address.go
anx-mschaefer: `0` is simply the zero-value which is omitted (via omitempty json option) which means that we don't have any limits on which version is requested.

An "enum" type would probably help :+1:. I think I'd leave out the `Dualstack`  value, as it might suggest that one of both types will be reserved. But it will only be one - either v4 or v6, depending on what is available. ..my comment was probably misleading. I'll try to clarify it.

Validation by the server should be sufficient, I think. But please let me know if you have a strong opinion on adding client side validation for this.

nachtjasmin (approved):

PR #391, opened 2024-07-29
Add missing anxcloud:identifier struct tags

nachtjasmin:
When working on our ake-controller-manager together with the usage of the go-anxcloud mock module, I've stumbled across this issue for the common.PartialResource.

This error is only occuring because the [makeObjectIdentifiable] function requires the `anxcloud:"identifier"` tag to be set. I haven't touched the code in the function itself, as any panic or error caused by it might break existing callers or usecases I haven't considered.

[makeObjectIdentifiable]: pkg/api/mock/utils.go

### Checklist

### Community Note

### DIFF
--- a/pkg/apis/clouddns/v1/zone_types.go
+++ b/pkg/apis/clouddns/v1/zone_types.go
@@ -4,7 +4,7 @@ import "time"
 
 type Revision struct {
 	CreatedAt  time.Time `json:"created_at"`
-	Identifier string    `json:"identifier"`
+	Identifier string    `json:"identifier" anxcloud:"identifier"`
 	ModifiedAt time.Time `json:"modified_at"`
 	Records    []Record  `json:"records"`
 	Serial     int       `json:"serial"`
--- a/pkg/apis/common/resource.go
+++ b/pkg/apis/common/resource.go
@@ -6,7 +6,7 @@ import (
 
 // PartialResource represents a linked resource
 type PartialResource struct {
-	Identifier string `json:"identifier,omitempty"`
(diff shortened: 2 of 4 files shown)


pkg/apis/clouddns/v1/zone_types.go
```diff
@@ -4,7 +4,7 @@ import "time"
 
 type Revision struct {
 	CreatedAt  time.Time `json:"created_at"`
-	Identifier string    `json:"identifier"`
+	Identifier string    `json:"identifier" anxcloud:"identifier"`
```
LittleFox94, line 7: The `Revision` type is not an `Object`, so it shouldn't be used by anything requiring this tag

Is there an Engine API to CRUD a specific revision? If so, adding this might still make sense

pkg/apis/common/resource.go
```diff
@@ -6,7 +6,7 @@ import (
 
 // PartialResource represents a linked resource
 type PartialResource struct {
-	Identifier string `json:"identifier,omitempty"`
+	Identifier string `json:"identifier,omitempty" anxcloud:"identifier"`
```
LittleFox94, line 9: PartialResource is not an Object and nothing should require this tag - I'll need more info what you tried to do and what happened please

pkg/apis/clouddns/v1/zone_types.go
nachtjasmin: welp, likely I applied the wrong solution to the problem. I think this PR can be closed then. ^^

PR #410, opened 2024-10-14
fix: make status field accessible

DrPsychick:
### Description

Make the `status` field of provisioning progress accessible.

### Checklist

### DIFF
--- a/pkg/vsphere/provisioning/progress/progress.go
+++ b/pkg/vsphere/provisioning/progress/progress.go
@@ -23,7 +23,7 @@ const (
 type Progress struct {
 	// TaskIdentifier is the identifier of the provisioning task.
 	TaskIdentifier string `json:"identifier"`
-	// Queued indicated that the task is waiting to be started-
+	// Queued indicates that the task is waiting to be started.
 	Queued bool `json:"queued"`
 	// Progress is the provisioning progress in percent (queuing not included).
 	Progress int `json:"progress"`
@@ -31,6 +31,8 @@ type Progress struct {
 	VMIdentifier string `json:"vm_identifier"`
 	// Errors encountered while provisioning.
 	Errors []string `json:"errors"`
+	// Status of the task.
+	Status int `json:"status"`
 }
 
 // ErrProgress is raised if a poll request completes but the result contains errors.
--- a/pkg/vsphere/provisioning/progress/progress.go
+++ b/pkg/vsphere/provisioning/progress/progress.go
@@ -17,10 +17,15 @@ const (
 	pathPrefix            = "/api/vsphere/v1/provisioning/progress.json"
 	pollInterval          = 5 * time.Second
 	progressCompleteValue = 100
-	statusFailed          = -1
-	statusSuccess         = 1
-	statusInProgress      = 2
-	statusCancelled       = 3
+
+	// StatusFailed indicates that the progress failed.
+	StatusFailed = -1
+	// StatusSuccess indicates that the progress succeeded.
+	StatusSuccess = 1
+	// StatusInProgress indicates that the progress is still ongoing.
+	StatusInProgress = 2
+	// StatusCancelled indicates that the progress has been cancelled.
+	StatusCancelled = 3
 )
 
 // Progress contains information regarding the provisioning of a VM returned by the API .
--- a/pkg/vsphere/provisioning/progress/progress.go
+++ b/pkg/vsphere/provisioning/progress/progress.go
@@ -19,14 +19,16 @@ const (
 	progressCompleteValue = 100
 
 	// StatusFailed indicates that the progress failed.
-	StatusFailed = -1
+	StatusFailed Status = -1
 	// StatusSuccess indicates that the progress succeeded.
-	StatusSuccess = 1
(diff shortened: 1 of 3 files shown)


89Q12 (approved):
Looks good, the changelog has a conflict that wants to be resolved but otherwise good work thank you :3

pkg/vsphere/provisioning/progress/progress.go
```diff
@@ -17,20 +17,31 @@ const (
...
+	// StatusCancelled indicates that the progress has been cancelled.
+	StatusCancelled = 3
 )
 
 // Progress contains information regarding the provisioning of a VM returned by the API .
 type Progress struct {
 	// TaskIdentifier is the identifier of the provisioning task.
 	TaskIdentifier string `json:"identifier"`
-	// Queued indicated that the task is waiting to be started-
+	// Queued indicates that the task is waiting to be started.
```
nachtjasmin, line 35: Good catch! :+1:

pkg/vsphere/provisioning/progress/progress.go
```diff
@@ -17,20 +17,31 @@ const (
...
 	Queued bool `json:"queued"`
 	// Progress is the provisioning progress in percent (queuing not included).
 	Progress int `json:"progress"`
 	// VMIdentifier of the new VM.
 	VMIdentifier string `json:"vm_identifier"`
 	// Errors encountered while provisioning.
 	Errors []string `json:"errors"`
+	// Status of the task.
+	Status int `json:"status"`
```
nachtjasmin, line 44: It probably makes sense to use a dedicated type for this instead of an generic `int`.

pkg/vsphere/provisioning/progress/progress_test.go
```diff
@@ -0,0 +1,109 @@
...
+
+		When("progress is cancelled", func() {
+			BeforeEach(func() {
+				identifier = "statusCancelledTest"
+				prepareGet(identifier, []string{}, StatusCancelled)
+			})
+
+			It("it returns status cancelled", func() {
+				Expect(requestErr).NotTo(HaveOccurred())
+				Expect(result).NotTo(BeNil())
```
nachtjasmin, line 89: Result is just an int and can never be nil, therefore the line can be removed.

```suggestion
```

pkg/vsphere/provisioning/progress/progress_test.go
```diff
@@ -0,0 +1,109 @@
...
+
+		When("progress has failed", func() {
+			BeforeEach(func() {
+				identifier = "statusFailedTest"
+				prepareGet(identifier, []string{}, StatusFailed)
+			})
+
+			It("it returns status failed", func() {
+				Expect(requestErr).NotTo(HaveOccurred())
+				Expect(result).NotTo(BeNil())
```
nachtjasmin, line 49:
```suggestion
```

pkg/vsphere/provisioning/progress/progress_test.go
```diff
@@ -0,0 +1,109 @@
...
+
+		When("progress has succeeded", func() {
+			BeforeEach(func() {
+				identifier = "statusSuccessTest"
+				prepareGet(identifier, []string{}, StatusSuccess)
+			})
+
+			It("it returns status success", func() {
+				Expect(requestErr).NotTo(HaveOccurred())
+				Expect(result).NotTo(BeNil())
```
nachtjasmin, line 62:
```suggestion
```

pkg/vsphere/provisioning/progress/progress_test.go
```diff
@@ -0,0 +1,109 @@
...
+
+		When("progress has errors", func() {
+			BeforeEach(func() {
+				identifier = "statusInProgressTest"
+				prepareGet(identifier, []string{"some error"}, StatusInProgress)
+			})
+
+			It("it returns an error", func() {
+				Expect(requestErr).To(HaveOccurred())
+				Expect(result).NotTo(BeNil())
```
nachtjasmin, line 75:
```suggestion
```

nachtjasmin (changes requested):
Overall: :+1:
Just some small things here and there, but those are adjusted quickly.

I'd also prefer it if we could add integration tests as well (yes, this includes creating a VM). But due to our time constraints, I'm also fine if covering this in a follow-up ticket.

pkg/vsphere/provisioning/progress/progress.go
DrPsychick: Yes, I thought about that too, but decided to keep it flexible and instead exposed the known statuses as constants.

I would put this change into the followup ticket mentioned for integration tests.

nachtjasmin: We wouldn't lose any of the flexibility if we convert it to a proper type, or am I missing something here?

pkg/vsphere/provisioning/progress/progress_test.go
nachtjasmin: Oh, my bad. We should remove the check nonetheless, since the zero value is always `!= nil`.

nachtjasmin (approved):

PR #419, opened 2024-11-06
SYSENG-1822: fix ResourceWithTag Identifier usage

DrPsychick:
### Description

ResourceWithTag expects the identifier of a resource that is to be tagged. The tag itself has its own identifier which is returned in the response from the Engine.

### Checklist

### Community Note


pkg/apis/lbaas/v1/e2e_connection_test.go
```diff
@@ -40,7 +40,7 @@ func unavailableServerConnectionCheck(url string) {
 			resp, err := http.Get(url)
 			g.Expect(err).NotTo(HaveOccurred())
 			g.Expect(resp.StatusCode).To(Equal(http.StatusServiceUnavailable))
-		}, 5*time.Second, 1*time.Second).Should(Succeed())
+		}, 60*time.Second, 5*time.Second).Should(Succeed())
```
nachtjasmin, line 43: This is not part of the ticket and should be moved into a new PR.

pkg/vlan/integration_test.go
```diff
@@ -41,10 +41,11 @@ var _ = Describe("vlan client", func() {
...
 			vlanID = summary.Identifier
-		})
-		DeferCleanup(func(ctx context.Context) {
-			err := api.Delete(ctx, vlanID)
-			Expect(err).NotTo(HaveOccurred())
+
+			DeferCleanup(func(ctx context.Context) {
+				err := api.Delete(ctx, vlanID)
+				Expect(err).NotTo(HaveOccurred())
+			})
```
nachtjasmin, line 48: Same goes for this change, please create a new PR for it.

pkg/apis/core/v1/resource_genclient.go
```diff
@@ -91,7 +90,7 @@ func (rwt ResourceWithTag) EndpointURL(ctx context.Context) (*url.URL, error) {
 		return nil, fmt.Errorf("%w: ResourceWithTag only support Create and Destroy operations", api.ErrOperationNotSupported)
 	}
 
-	return url.Parse(fmt.Sprintf("/api/core/v1/resource.json/%v/tags/%v", rwt.Identifier, rwt.Tag))
+	return url.Parse(fmt.Sprintf("/api/core/v1/resource.json/%v/tags/%v", rwt.ResourceIdentifier, rwt.Tag))
```
nachtjasmin, line 93: :warning: This would break *all* existing usages for this resource!

The ticket you've created says:

> When using the mock, this causes a resource object to be replaced in the mock data with the ResourceWithTag object because they have the same identifier.

It's clear that this is a problem inside the implementation of the mock, not of the actual API. There are two solutions to this problem:

1. Either provide a way that sets the `ResourceIdentifier` in a backwards compatible way for existing users *or*
2. fix the wrong behavior in the mock itself.

My preference would be option 2, but I'm fine with option 1 as well.

DrPsychick: We agreed on adjusting the code so it is backwards compatible, using `Identifier` when no `ResourceIdentifier` is available.

DrPsychick: Thanks for that feedback, I additionally found an issue with the tests and fixed it as well.

pkg/apis/lbaas/v1/e2e_connection_test.go
DrPsychick: This is part of this PR as it only affects/fixes integration tests.

pkg/vlan/integration_test.go
DrPsychick: This is part of this PR as it only affects/fixes integration tests.

pkg/api/mock/mock_api_implementation.go
```diff
@@ -296,6 +296,11 @@ func (a *mockAPI) FakeExisting(o types.Object, tags ...string) string {
 	a.dataMu.Lock()
 	defer a.dataMu.Unlock()
 
+	// SYSENG-1822: faking ResourceWithTag blindly overwrites object with same identifier.
+	if _, ok := o.(*corev1.ResourceWithTag); ok {
+		panic("Cannot fake ResourceWithTag, pass tags to the object instead.")
```
nachtjasmin, line 301: Little suggestion for a better wording:

```suggestion
		panic("mock: cannot fake ResourceWithTag, pass tags to the object or use the FakeExisting method")
```

nachtjasmin (changes requested):
One minor suggestion but then we're ready to merge this! Thanks a lot!!

nachtjasmin (approved):

